Issue #589, opened 2018-10-10
Code Nodes do not display empty lines


qwertynerd97:
Although empty lines can be put into the textGroup for a Code node, they will not display in the viewer.

PR #590, opened 2018-10-10
Fix code wrapping and empty lines

### LINKED ISSUE #588: Really Long Code Overflows Its Container
A really long line of code will simply extend to the right indefinitely, making the whole page look lopsided.
Wrapping the line is probably not the solution we want, as a new line has a specific meaning in code.
On most websites, the code block itself would become scrollable on overflow.

zachberry:
Resolves #588 and #589 

* Fixes code node having wrong style (Was using single-text)
* Fixes blank lines not showing up
* Styles code in a box which now scrolls horizontally
* Updates tab length to be two spaces instead of four
* Reduces vertical margins (Code elements were pushed too far away from any element above or below - which became obvious when the code was put inside a box)


gitnix:
Looks good.

Issue #608, opened 2018-10-26
WYSIWYG: Cancelling out of Rename Module sets the name of the draft to the ID


zachberry:
1. Click Rename Module
2. Click Cancel

Result: Document title in nav bar shows up as the draft ID.

PR #615, opened 2018-10-26
Fixed module renaming issue

qwertynerd97:
Set up renaming so that canceling the rename keeps the old name and pressing "OK" with a blank name field renames it to a placeholder name.

Resolves https://github.com/ucfopen/Obojobo/issues/608

### DIFF
--- a/packages/obojobo-document-engine/src/scripts/oboeditor/components/editor-nav.js
+++ b/packages/obojobo-document-engine/src/scripts/oboeditor/components/editor-nav.js
@@ -51,10 +51,13 @@ class EditorNav extends React.Component {
 	}
 
 	renamePage(pageId) {
-		const label = window.prompt('Enter the new title:')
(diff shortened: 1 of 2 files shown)


packages/obojobo-document-engine/src/scripts/oboeditor/components/editor-nav.js
```diff
@@ -51,10 +51,28 @@ class EditorNav extends React.Component {
...
 	}
 
+	renameModule(moduleId) {
+		let label = window.prompt('Enter the new title:')
+
+		// null means the user canceled without changing the value
+		if (label === null) return
+
+		// If the module name is empty, provide a default value
+		if (!label) label = '(Unnamed Module)'
```
gitnix: same here--
```
// tests if the string is empty or only whitespace
if (!label || /\s/.test(label)) label = '(Unnamed Module)'

Issue #612, opened 2018-10-26
WYSIWYG: Deleting all the content in a page causes that page to be stuck blank


zachberry:
* Delete all the content on a page
* Result: The cursor jumps to the upper left and then that page gets stuck into a blank state

PR #617, opened 2018-10-26
Fixed pages that were stuck as blank

qwertynerd97:
Added a schema to the top-level document that ensures all pages have at least one node in them.  This also sets Text nodes up as the default node; if a user deletes the last node in a page, the page will insert a new, empty text node.

Resolves https://github.com/ucfopen/Obojobo/issues/612

### DIFF
--- a/packages/obojobo-document-engine/src/scripts/oboeditor/plugins/editor-schema.js
+++ b/packages/obojobo-document-engine/src/scripts/oboeditor/plugins/editor-schema.js
@@ -1,0 +1,57 @@
+import { CHILD_REQUIRED, CHILD_TYPE_INVALID } from 'slate-schema-violations'
+
+// Default node type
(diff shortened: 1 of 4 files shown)


packages/obojobo-document-engine/src/scripts/oboeditor/plugins/editor-schema.js
```diff
@@ -0,0 +1,57 @@
...
+]
+
+const matcher = nodes.map(chunk => {
+	return { type: chunk }
+})
+
+const EditorSchema = {
+	schema: {
+		document: {
+			nodes: [{ match: matcher, min: 1 }],
```
gitnix: since matcher doesn't do anything fancy, would probably just opt for 
`nodes: [{ match: nodes.map(n => {type: n}), min: 1 }]`

Issue #613, opened 2018-10-26
WYSIWYG: HTML tag causes Maximum Stack Exceeded errors when put in Assessment


zachberry:
* Add `<HTML html="<hr />" />` either in the Assessment landing page or in a scoreActions page.
* Result: Editor doesn't render and seems and throws a max stack exceeded error

When included in a content page the editor doesn't crash, however the page simply shows the text `<hr />` instead of an hr element.

PR #619, opened 2018-10-29
Issue/613 html rendering

qwertynerd97:
Adds more clear rendering for HTML nodes, and allows HTML nodes to be rendered in assessments and questions.
HTML nodes now render similarly to code nodes in the editor, and are label as being HTML (so they are not mistaken for code nodes).  Although they don't have all the same capabilities of a Code node, they do allow enters and indents to enable users to format their HTML in a more readable way.

Resolves https://github.com/ucfopen/Obojobo/issues/613


qwertynerd97:
The backspacing behavior is standard across all (text-based) nodes and is similar to the behaviors of Word documents.

Issue #280, opened 2018-09-07
Change course navigation LTI config to use 'admin' instead of 'members'


iturgeon:
<a href="https://github.com/zachberry"><img src="https://avatars2.githubusercontent.com/u/73479?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zachberry](https://github.com/zachberry)**
_Friday Aug 17, 2018 at 19:18 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Next/issues/282_

----

This might allow teachers to see the item but not students

zachberry:
Pull request: #622

PR #622, opened 2018-10-30
Updates course_navigation/visibility from members to admins

zachberry:
This will hide this from students - See https://canvas.instructure.com/doc/api/file.navigation_tools.html)


gitnix:
resolves #280

Issue #649, opened 2018-12-14
Move certain properties registered through the model store to adapter


gitnix:
For example, 
```
getNavItem(model) {
...
}
```
could be moved to adapter. Also the `variables`  object in Sections/Assessment could become something like
```
getTextForVariable(name, model, props) {
    ... 
}
```
on adapter

PR #650, opened 2018-12-14
Viewer/Editor separation

gitnix:
**WIP**

* removed `obojobo.json`
* removed `obo_get_installed_modules`
* now using a root-level `obojobo.js` file
This file is simple for now, but could eventually become more complex.
(It would probably eventually become a json file)
Did not like the complicated nature of obo_get_installed_modules (scouring the directories for `obojobo.json` files) Instead, the obojobo file must now be defined in the root directory. More complicated behavior can be added in the future as necessary, such as supporting multiple obojobo files.

* removed webpack files for docEngine and express projects.
There is now a single webpack file in the root directory.

* updated package.json files
    * removed unnecessary scripts
    * removed unnecessary dependencies
    * updated all dependencies
    * now using new @babel instead of babel (updated babelrc)

* removed unneeded yarn.lock files for all packages
(only the top-level yarn.lock is used)

* Renamed `Common.Store` to `Common.Registry`
This aligns with registering models to a Registry.
This also clears confusion when dealing with flux store vs common store
The editor registers models to Common.Registry
We could eventually have a Viewer Registry and an Editor Registry
Currently `getNavItem` code is duplicated (on both viewer and editor) model registration (which could be changed with https://github.com/ucfopen/Obojobo/issues/649)

Since the editor and viewer build are now separated, Editor components need to import `viewer-component.scss` to be visually presentable. Eventually all editor specific css will be pulled out to an `editor-component.scss` along with all css common to both editor and viewer (possibly `component.scss`?).
note: css for attempt review correct/incorrect answers need to be adjusted

Besides the components themselves, the editor needs to import much of the non-chunk viewer css for the app to display properly. This css should eventually be reworked to account for the common needs of the editor and viewer.

* Added match for `ObojoboDraft.Sections.Assessment` in `editor-schema`
which is needed to render an assessment section

* Updated `express_register_chunks` to use `obojobo.js`

* Began removing old editor code

will need to update tests

known bugs:
when switching to an assessment in editor- getting `node with duplicate key`


packages/obojobo-document-engine/ObojoboDraft/Chunks/Question/editor.js
```diff
@@ -23,14 +23,34 @@ const plugins = {
...
+Common.Registry.registerModel('ObojoboDraft.Chunks.Question', {
 	name: 'Question',
 	icon: Icon,
 	isInsertable: true,
 	insertJSON: emptyNode,
 	slateToObo: Converter.slateToObo,
 	oboToSlate: Converter.oboToSlate,
-	plugins
+	plugins,
+	getNavItem(model) {
```
qwertynerd97: This is the same question as the heading: since we are not displaying this in the navigation, do we really want to register the nav item?

PR #654, opened 2019-01-08
Issue/641 update test object to use math equations for example questions

rmanbaird:


gitnix:
Looks good. The coverage file isn't needed, but it doesn't appear to be in the `.gitignore` of the xml project. I would add `coverage` directory to `.gitignore` and remove the generated coverage file.

PR #701, opened 2019-02-07
Adds npm publish to Travis, lerna, and package.json info

iturgeon:
* Travis: Removes cache for node_modules in packages
* Travis: Adds stages for each package to publish to npm
* Adds lerna so we can use `lerna version` scripts for easy updating of the workspace versions
* Adds repo, license, and homepage to all package.json files


package.json
```diff
@@ -13,7 +16,8 @@
   },
   "devDependencies": {
     "concurrently": "^3.5.0",
-    "husky": "^0.13.3"
+    "husky": "^0.13.3",
+    "lerna": "^3.10.7"
```
gitnix: lerna intended?

iturgeon: I mentioned this in person - but lerna is added here so we use the version command to update the version numbers in all the package.json files in each workspace.

https://github.com/lerna/lerna/tree/master/commands/version

.travis.yml
```diff
@@ -6,8 +6,56 @@ cache:
...
+      node_js: '10.4.1'
+      # move into the desired package and back out for deploying
+      before_deploy:
+        - cd packages/obojobo-express
+      after_deploy:
+        - cd ../../
+       # deploy this package to npm
+      deploy:
+        provider: npm
+        email: iturgeon+npm@gmail.com
```
zachberry: Alright to use personal email address? Do we need to use a UCF address?

iturgeon: this email is the email of the npm user that is deploying. There's plusses and minuses to going with an organization account vs an individual account.

This option is a few less hoops now for a few extra hoops later should I leave.   Multiple people have npm ownership over the repo, so any of the other owners can setup a shared secret with travis and use issue a PR to update this.

If you feel strongly about it I'd suggest splitting it off to an issue and going forward w/ this as is

zachberry: Made #753

zachberry:
Discussed IRL that we're gonna to push this to dev/5 instead of master, since dev/5 is close to being wrapped up

PR #741, opened 2019-03-01
Allow nav to scroll

### LINKED ISSUE #709: Can't scroll through side nav (In viewer and editor)
Based on a comment by @gitnix - Looks like you can't scroll the side nav. If there are several items in the nav it's impossible to navigate to them on a height-restricted screen.

zachberry:
Fixes #709

### DIFF
--- a/packages/obojobo-document-engine/src/scripts/oboeditor/components/sub-menu.scss
+++ b/packages/obojobo-document-engine/src/scripts/oboeditor/components/sub-menu.scss
@@ -5,7 +5,7 @@
 
 	.dropdown {
 		position: absolute;
-		right: 0;
+		right: -1em;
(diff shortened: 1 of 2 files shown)


zachberry:
@samuel-belcastro Same thing happened to me - it's because on default OS X hides the scrollbars unless you enable them all the time (System Settings -> General -> Show Scroll bars). Only when 'Always' is selected does the layout change.

zachberry:
Last minor update should address @qwertynerd97 's issue. When scrollbars are always displayed, scrollbars are needed and the small sidebar nav is shown the arrows touch against the sidebar, which is not ideal, but this is a compromise to make it work for all options. In the future I see us tweaking the arrow.

PR #744, opened 2019-03-07
Revert 714 issue/690 update slate

iturgeon:


iturgeon:
Canceling - the revert didn't have the effect we were expecting

Issue #724, opened 2019-02-19
Copy and Paste in Editor


qwertynerd97:
Copying text from the editor and then pasting it back into the editor fails to paste the content.  This may be due to how Slate represents data; we can probably override this with a plugin.

qwertynerd97:
Notes on copy-pasting with merging:
We have 4 types of nodes: 
- Void nodes (contain no text in the editor, only modal properties)
- Single nodes (contain only one line of text)
- Multi nodes (contain one or more lines of text)
- Parent nodes (contain other nodes, but no text of their own)

Therefore, there are 16 types of paste operations:
1. Void into Void - Insert pasted node at end of receiving node
2. Void into single - Insert pasted node into single node, split single node if pasted in the middle
3. Void into multi - Insert pasted node into multi node, split multi node if pasted in the middle
4. Void into parent - Does not actually exist (pasting into parent is actually pasting into void, single, or multi inside parent)
5. Single into void - Insert single at end of void node
6. Single into Single - Paste only the text into the receiving node
7. Single into multi - Paste only the text into the receiving node
8. Single into Parent - Cannot happen, see # 4
9. Multi into void - Insert multi at end of void
10. Multi into single - If at end of single: insert the first line of multi into single, and then create a multi after single with rest of lines.  If at start of single, insert the last line of multi and then create multi before single with rest of lines.  If in the middle of single and only one line in multi, add single line of text.  If in middle of single and multiple lines in multi, split single, add the first line to the first single, add the last line to the last single, and then create a new multi in between the singles to hold any extra lines
11. Multi into multi - convert lines to correct type and insert, pre- or post-pending as necessary
12. Multi into parent - Cannot happen, see # 4
13. Parent into Void - Insert pasted node at end of receiving node
14. Parent into single - Insert pasted node into single node, split single node if pasted in the middle
15. Parent into multi - Insert pasted node into multi node, split multi node if pasted in the middle
16. Parent into parent - Cannot happen, see # 4

These make the most sense if only pasting one node; if pasting multiple nodes, use this decision tree to paste the first node, and then simply insert successive nodes.

Current Obojobo nodes and their (potential) types:
ActionButton - Parent node containing Single node label + void node actions (pasting only actions would create an action button with a blank label)
Break - void node
Code - multi node
Figure - Parent node containing void image + single node caption (pasting only image would create a Figure with a blank caption, pasting into the image would insert before instead of after)
Heading - single node
[...]

PR #746, opened 2019-03-07
Copy and Paste

qwertynerd97:
Improves paste features in the editor.
Users can now copy Items in the editor and then paste them back in, without having to go through a secondary text editor.  
Users can copy the content of one MC Choice and then paste it into another
Users can copy and paste within the Assessment section
Upgraded to a minor fix of Slate, so users can now cut within the OboEditor.
Adds the supportsChildren parameter to OboNode registration; OboNodes that support children will reduce the tree structure to leaf-most OboNodes when pasting.  This prevents users from pasting Questions inside themselves and causing errors.

Partially implements a solution to https://github.com/ucfopen/Obojobo/issues/724
Does not include merging paste


packages/obojobo-document-engine/ObojoboDraft/Chunks/Table/schema.js
```diff
@@ -21,6 +21,11 @@ const schema = {
 				const header = index === 0 && node.data.get('content').header
 				switch (error.code) {
 					case CHILD_TYPE_INVALID: {
+						// Deal with Paste insertion of top-level nodes
```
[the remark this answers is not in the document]
zachberry: Agree

packages/obojobo-document-engine/__tests__/ObojoboDraft/Chunks/Code/editor.test.js
```diff
@@ -424,6 +420,21 @@ describe('Code editor', () => {
...
 	})
 
+	test('plugins.schema.normalize fixes invalid oboeditor.component in code', () => {
+		const editor = {
+			unwrapNodeByKey: jest.fn()
+		}
+
+		Code.plugins.schema.blocks[CODE_NODE].normalize(editor, {
+			code: CHILD_TYPE_INVALID,
+			node: { nodes: { size: 10 } },
```
zachberry: What does the "size 10" bit mean here?

qwertynerd97:
Talked about IRL: 
Right now, pasting Obojobo Editor nodes into other nodes creates a new node every time.  In the future, we would like to be able to paste a node and have it merge into the existing node.  Currently, the way Slate's command system is set up, this is very difficult.  Slate changed their command structure in 0.45 (as well as breaking a couple of other things), so when they release 0.46 we might want to upgrade and take another look at how we are doing our pasting.

zachberry:
User comments who experienced a bug with the current editor. Want to test to see if these changes fix these issues:

```
Running into a bug with the editor
When I select the title of a page and hit CTrl + V trying to paste some text, the editor crashes
this seems to happen only when I select the entire title, and hit Ctrl + V
and often happens on the second attempt
on the first attempt, the text I copied will be pasted in the middle of the existing text, instead of replacing it.
then on the second attempt the whole thing crashed
re-freshing the page seems to be able to restore the editor
and selecting part of the title doesn't cause the bug

Only affects headings of any level (edited) 
```

zachberry:
I tried the above ^ - it works. The behavior has changed so that bug doesn't occur. I think we still want to move towards the behavior mentioned in the last post by @qwertynerd97

packages/obonode/obojobo-sections-assessment/schema.js
```diff
@@ -25,6 +25,7 @@ const schema = {
 				{ match: [{ type: RUBRIC_NODE }] }
 			],
 			normalize: (editor, error) => {
+				console.log('assessment', error)
```
zachberry: Left in console
```suggestion
```

zachberry:
Spending some more time trying to break things. I can't figure out a consistent pattern yet, but these are the things I've found:

* [ ] - Sometimes using Ctrl+A to select all then pressing delete doesn't delete all the content. Instead it deletes a single character where the cursor was.
* [ ] - Sometimes you can't use Ctrl+A, instead the selection flashes then disappears. When this happens you need to change pages to restore normal Ctrl+A functionality.
* [ ] - Pasting in complicated content, like big HTML pages or large Obojobo pages, still can fire errors in the editor and the past doesn't go through. So far however these haven't killed the editor, it seems to be able to continue working.

I don't want to block these enhancements further so I'm going to move the above into a new issue, but I wanted to mention it here regardless.

Issue #718, opened 2019-02-15
JSON to XML parser


zachberry:
In order to have a 'source view' with the editor we need to "reverse" parse the JSON back into XML.

PR #758, opened 2019-03-22
Add XML to JSON warning for wysiwyg editor

### LINKED ISSUE #757: Add warning when editing an XML document in the WYSIWYG editor that the document will be "converted" to JSON
Authors probably won't be aware that editing a document in the WYSIWYG editor will "lose" their XML and the document will now be JSON only. I'm thinking we at least add a JS alert warning them that this will happen.

zachberry:
Fixes #757

Adds a JS confirm prompt if attempting to edit an XML document in the WYSIWYG editor


gitnix:
We should probably also create a copy of the document and tell the user something along the lines of `A Backup XML draft will be created`.

zachberry:
Talked with @gitnix IRL, deciding to defer the copy feature. Once #718 is built this won't be necessary, and since that's on the near horizon thinking it may not be necessary to spend dev time on it.

PR #796, opened 2019-04-05
Fix cross browser styles

qwertynerd97:
Fixes styles so that everything in the editor is consistent between browsers (Tested in Chrome, Firefox, and Safari)
Fixed bug with Link modal


samuel-belcastro:
In regards to @gitnix comment above:
> text goes over pop-up menu

Looks like this css rule is the culprit:
```
.editor--components--nav .link.is-selected {
    z-index: 9999;
}
```

Only selected nav items are showing up over the pop up nav controls in the editor.

Issue #797, opened 2019-04-08
Draft Manager in IE11 crashes due to arrow function syntax


zachberry:
There's some IE11 unfriendly editor code that needs refactoring (since this file isn't babel-transformed)

zachberry:
We moved to Babel this so this shouldn't be an issue anymore

PR #798, opened 2019-04-09
Issue/797 draft manager javascript split

iturgeon:
* Moves most javascript out of editor_picker.ejs into draftmanager.js
* Found a way to get webpack to compile a single file without 
* test coverage will drop on docengine because of the new file
* need some feedback on where to place the draftmanager.js
Solves #797


iturgeon:
worth noting - this only passes tests because of a jest bug #799 if tests are run in the sub package, the coverage fails.

Waiting for everyone's thoughts before looking at that.

PR #813, opened 2019-04-17
adds lti launch routes in dev environment

iturgeon:
adds the following routes when webpack-dev-server is active:

* /dev (links to all the things below)
* /routes
* /lti/dev/launch/course_navigation
* /lti/dev/launch/view
* /lti/dev/launch/resource_selection
* /lti/dev/return/resource_selection


packages/app/obojobo-express/obo_express_dev.js
```diff
@@ -0,0 +1,181 @@
...
+			.sort((a,b) => a.path.localeCompare(b.path))
+
+		res.json(simplifiedEndpoints)
+	})
+
+	// builds a valid course navigation lti launch and submits it
+	app.get('/lti/dev/launch/course_navigation', (req, res) => {
+		const method = 'POST'
+		const endpoint = `${baseUrl(req)}/lti/canvas/course_navigation`
+		const params = {
```
gitnix: would suggest making a common params object that contains user info and spreading over that inside this and subsequent route params

packages/app/obojobo-express/webpack.config.js
```diff
@@ -44,6 +45,7 @@ module.exports =
 				watchOptions: {
 					ignored: '/node_modules/'
 				},
+				setup: ObojoboExpressDev,
```
gitnix: The setup option is deprecated. Suggest moving into `before` function.

zachberry: #busted

packages/app/obojobo-express/obo_express_dev.js
zachberry: Agreed

iturgeon:
I'd argue for that if their values would ever be used anywhere else.  IMO, they have no use anywhere but in this page - and they are unlikely to be something you want to change given the use case for these pages.

Issue #795, opened 2019-04-05
Visual Editor Mockup


zachberry:
A placeholder type of issue to keep track to the improvements/polish we want to add to the visual editor. This visual editor would become both the WYSIWYG and the older XML editor (as in they'd be combined into one interface).

![Screenshot 2019-04-05 14 13 35](https://user-images.githubusercontent.com/73479/55648316-fc939580-57ad-11e9-9207-3c808fee53e5.png)

Current thought is that the UI would remain purple for actions but any in-line actions that appear in the content would be blue. Not 100% sold on this idea yet. Menu on the right would let you toggle between normal editing and previewing a tablet and phone view (to let authors see how their content roughly displays on smaller screens). File menu on the top would give you access to all functions while common functions would be on the toolbar below. The + button next to 'Insert Item...' would be a hotbar allowing you to add commonly-used chunks (like the existing Table and Question icons). The far right menu with the paintbrush would allow you to change themes. The gear next to that would be a module-level options menu (Perhaps editing things such as the 'start' property).

Editing latex:

![Untitled-1](https://user-images.githubusercontent.com/73479/55743498-7753ef00-5a00-11e9-876a-f88b095c1917.png)

Changing a heading:

![Screenshot 2019-04-05 14 14 01](https://user-images.githubusercontent.com/73479/55648516-87749000-57ae-11e9-8796-4c4849b2ab73.png)

Code view (A 2:2 split would be cool! But not necessary)

![Screenshot 2019-04-05 14 14 09](https://user-images.githubusercontent.com/73479/55648544-99563300-57ae-11e9-8e05-079ee0355b51.png)

The circle buttons with the `...` would give you access to generic node properties:

![Screenshot 2019-04-05 14 14 20](https://user-images.githubusercontent.com/73479/55648573-abd06c80-57ae-11e9-9e31-997a74f94689.png)

They would only show up when that node was focused in on. This would let you perform the more dynamic 'plumbing' of the document. The lightning icons could represent any node which has triggers on it. That way it is showing you that there is more to that item then you can see at face value.

Perhaps from that `...` menu (or by some other means) you could edit the XML of a single node:

![Screenshot 2019-04-05 14 14 36](https://user-images.githubusercontent.com/73479/55648678-f05c0800-57ae-11e9-81e0-519fed83d96e.png)

[...]

PR #814, opened 2019-04-18
Visual Editor Redesign

qwertynerd97:
Re-designs the Visual Editor to more closely match https://github.com/ucfopen/Obojobo/issues/795.
Moves the Mark toolbar to the top of the page
Adds a file toolbar
Creates a distinct layout to visually differentiate the editor from the viewer
Breaks out mark functions for better modularity


qwertynerd97:
Notes: I have the toolbars all set up, but only partially working.  The content toolbar allows the user to add bold, italic, align, etc, and is fully functional.  The file toolbar handles all the file functions, as well as the content functions.  Within the file toolbar, the File menu works, although none of the functions have shortcut keys.  None of the other menus work currently.

iturgeon:
When i click copy, the page being copied has the same title as the copy: `page 3 (copy)`.

I'm guessing we need to deep clone the object (Registry's new `cloneBlankNode` method). Perhaps we need to deal with new object ids too?

iturgeon:
This PR is replaced by #989

PR #823, opened 2019-04-22
allow json string in env vars

iturgeon:
env variables can now be valid json strings.  This was needed to allow multiple keys in the lti config without replacing the config.json file

Given the following env:
`ANY_CONFIG_JSON={"host":"mock-host","port":999}`

combined with the following config/sample.json file:

```javascript
{
    development: { ENV: 'ANY_CONFIG_JSON'}
}
```

will expand to:

```javascript
{
    development: {
        host: "mock-host",
        port: 999
    }
}
```

If the env var's name ends with `_JSON`, the code will log an error if the variable is not able to be parsed with `json.parse`. 

This check is only done when the name ends with `_JSON` to:
1. encourage use of a naming convention
2. avoid problems in assuming values that start with `{` or `[` are supposed to be json

## Important

This does change the default LTI configuration for production.

It also changes the Heroku install steps, which will now provide a static default key/secret pair instead of an automatically generated random key with a static key.

Now you'll need to provide a secret otherwise it will be a little dangerous to use with the default password that is provided.  We need to be sure to document the importance of that variable in our Heroku setup documentation.

### DIFF
--- a/packages/app/obojobo-express/config.js
+++ b/packages/app/obojobo-express/config.js
@@ -36,16 +36,15 @@ const isStringJSON = (name, string) => {
 	}
 }
 
-const replaceENVsInObject = configObject => {
-	const rawJson = JSON.stringify(configObject) // convert back to string
-
+const replaceENVsInJson = originalJson => {
+	// const originalJson = JSON.stringify(configObject) // convert back to string
 	// replace any "ENV": "CONFIG_VAR" settings with
 	// values from procesess.env
 	const pattern = /\{\s*"ENV"\s*?:\s*?"(.*?)"\s*\}/gi
 	let result
 
-	let replacedJson = rawJson
-	while ((result = pattern.exec(rawJson))) {
+	let replacedJson = originalJson
+	while ((result = pattern.exec(originalJson))) {
 		const envVar = result[1]
 		if (!process.env[envVar]) {
 			throw new Error(`Expected ENV var ${envVar} is not set`)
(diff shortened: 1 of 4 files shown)


iturgeon:
If we add the ability to recurse the env var replacement method - We would be able to still use Heroku's random secret generator, so I'll put a hold on this till that is ready.

packages/app/obojobo-express/config.js
```diff
@@ -24,30 +24,47 @@ if (process.env.DATABASE_URL) {
 	process.env.DB_PORT = dburl.port
 }
 
-const replaceENVsInObject = configObject => {
-	const rawJson = JSON.stringify(configObject) // convert back to string
+const isStringJSON = (name, string) => {
+	try{
```
gitnix: Regarding comment `"avoid problems in assuming values that start with { or [ are supposed to be json"` This automatically parses string as JSON here

gitnix:
logger is not printing to console. However, using `console.log` in place of `logger` gives expected behavior.

iturgeon:
It's logging out fine here:

```› yarn dev
yarn run v1.15.2
$ cd node_modules/obojobo-express && yarn dev
$ DEBUG=obojobo_server:error,obojobo_server:warn,obojobo_server:info IS_WEBPACK=true nodemon ./node_modules/.bin/webpack-dev-server
[nodemon] 1.18.10
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: /Users/iturgeon/git/Obojobo/packages/app/obojobo-express/**/* ../obojobo-document-engine/server
[nodemon] starting `node ./node_modules/.bin/webpack-dev-server`
OboNode client scripts to build | viewer: 19, editor: 19
>>  obojobo_server:error Error loading config file: /Users/iturgeon/git/Obojobo/packages/app/obojobo-express/config/lti.json +0ms
>>  obojobo_server:error Error: Expected ENV var OBO_LTI_KEYS_JSON is not set +1ms
```

zachberry:
Also seeing error logs as expected

PR #832, opened 2019-05-10
Assessment option in visual editor to lock nav for attempts in progress

### LINKED ISSUE #817: Navigation control not editable in visual editor
There doesn't seem to be a way to add actions such as nav:lock and nav:unlock triggered by entering or exiting a page or assessment, using the visual editor. This is critical for creating useful modules via the visual editor.

gitnix:
If `content.lockAssessment` is set in Assessment Settings, `onStartAttempt` and `onEndAttempt` triggers will be added to assessment.

Closes #817


packages/obonode/obojobo-sections-assessment/adapter.test.js
```diff
@@ -33,6 +33,34 @@ describe('ObojoboDraft.Sections.Assessment adapter', () => {
...
 	})
 
+	test('construct sets triggers when lockAssessment set', () => {
+		const attrs = {
+			content: { lockAssessment: true }
+		}
+		const model = new OboModel(attrs)
+		model.triggers = []
+		AssessmentAdapter.construct(model, attrs)
+		// expect(model.modelState).toMatchSnapshot()
```
zachberry:
```suggestion
```

packages/obonode/obojobo-sections-assessment/adapter.js
```diff
@@ -16,10 +16,29 @@ const Adapter = {
 			model.modelState.rubric = new AssessmentRubric()
 		}
 
-		model.setStateProp(
-			'attempts',
-			Infinity,
-			p => (('' + p).toLowerCase() === 'unlimited' ? Infinity : parseInt(p, 10))
+		if (attrs && attrs.content && attrs.content.lockAssessment) {
```
zachberry: I'm a little weary of adding a new attribute to content for this, since it sorta covers the same use case as the triggers and kind of modifies the document to add them in. I think I'd rather the WYSIWYG editor simply adds or removes these triggers.

packages/app/obojobo-document-engine/src/scripts/common/util/trigger-util.js
```diff
@@ -0,0 +1,96 @@
...
+	for (const type in usesMap) {
+		usesMap[type] = false
+	}
+
+	for (const trigger of triggers) {
+		let actions = triggerMap[trigger.type]
+		if (actions) {
+			if (!trigger.actions) trigger.actions = []
+			// if just an object convert to array
+			if (actions.constructor.name === 'Object') actions = [actions]
```
zachberry: I like using typeof, like `typeof actions === 'object'` - would that work or is constructor.name considered safer/more accurate?

zachberry: Actually, what about...

```suggestion
			actions = [].concat(actions)
```

I think that should work both if actions is already an array or if it's not

packages/app/obojobo-document-engine/src/scripts/common/util/trigger-util.js
```diff
@@ -0,0 +1,96 @@
...
+				// if post-filtered actions is empty add trigger to removal array
+				if (trigger.actions.length === 0) triggersToRemove.push(index)
+			}
+		}
+	})
+	// remove triggers with no actions
+	triggersToRemove.forEach((current, idx) => {
+		triggers.splice(current - idx, 1)
+	})
+}
```
zachberry: This code is doing some complex stuff and is still pretty hard to parse. I don't know if it can be simplified that much, so just adding lots more comments I think would help.

iturgeon: Perhaps this method shouldn't mutate it's arguments in place?  Would this work returning a new trigger? 
Doing so would also allow for some restructuring of the implementation - build a new array only with the things that pass and skipping things that need to be removed

gitnix: functions have been updated to return new triggers

PR #849, opened 2019-06-04
rename add Break to add Horizontal Line

vutoan1245:
Resolve issue #731


packages/obonode/obojobo-chunks-break/editor.js
```diff
@@ -22,7 +22,7 @@ const plugins = {
 }
 
 Common.Registry.registerModel('ObojoboDraft.Chunks.Break', {
-	name: 'Break',
+	name: 'Horiozonal Line',
```
zachberry:
```suggestion
	name: 'Horizontal Line',
```

PR #855, opened 2019-06-11
Add `JSON to XML` parser

vutoan1245:
Create a new folder `obojobo-document-json-parser` in `app`.
`json-to-xml-parser.js` export one function. This function takes in a javascript object and return a string in xml format.

Resolve issue #718


vutoan1245:
@zachberry I just realized that my account is not connected to these commits. I will close this pull request and make a new one.

PR #860, opened 2019-06-17
Issue/718 add `json to xml` parser

vutoan1245:
Create a new folder `obojobo-document-json-parser` in `app`.

`json-to-xml-parser.js` exports one function. This function takes in a javascript object and return a string in xml format.

Resolve issue #718


packages/app/obojobo-document-json-parser/src/text/chunk-style-list.js
```diff
@@ -0,0 +1,294 @@
+const StyleType = require('./style-type')
```
zachberry: I think this file is included twice

zachberry: (And some of the others below)

zachberry:
Looking really awesome so far. What I found in playing with it:

* `<solution>` in questions should generate `Page` elements, but instead only `Text` elements are created.
* Need to add a CLI way to run the parser. I'm thinking add a `json2xml.js` file in a `bin` directory, then add a script in the package.json to run it. You can see an example of this idea in the `xml2draft.js` file in the xml parser package. It should print usage information like the `xml2draft` script.

packages/app/obojobo-document-json-parser/__tests__/obo_node_json/mathequation.json
```diff
@@ -0,0 +1,9 @@
+{
+	"id": "66ae62f9-7cde-4a7f-9164-9245e9be1b1e",
+	"type": "ObojoboDraft.Chunks.MathEquation",
+	"content": {
+		"label": "1.1",
+		"latex": "x_1 = x_0 + \\Delta x"
```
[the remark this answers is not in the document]
vutoan1245: I will update

packages/app/obojobo-document-json-parser/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "dependencies": {
+  },
+  "devDependencies": {
+    "lint-staged": "^6.1.1"
+  },
+  "name": "json-to-xml-parser",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
vutoan1245: I updated to "version": "4.0.1"

packages/app/obojobo-document-json-parser/src/common/mockdom/mock-element.js
```diff
@@ -0,0 +1,51 @@
+class MockElement {
```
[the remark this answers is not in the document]
vutoan1245: I will probably need somebody to help me with this

packages/app/obojobo-document-json-parser/src/common/text/styleable-text-component.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
```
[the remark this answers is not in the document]
vutoan1245: I don't think we need it, I will delete it.

packages/app/obojobo-document-json-parser/src/text/chunk-style-list.js
vutoan1245: I included it multiple times, but in different files. Not sure if I miss something.

packages/app/obojobo-document-json-parser/src/common/mockdom/mock-element.js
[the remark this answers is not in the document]
vutoan1245: The current solution is to change all `import` to `require` in document engine. It seems to work fine.

packages/app/obojobo-document-json-parser/__tests__/jsonToXml.test.js
```diff
@@ -0,0 +1,16 @@
...
+const filterFileType = arr =>
+	arr.filter(s => s.includes('.json')).map(s => s.substring(0, s.length - 4))
+
+const nodes = filterFileType(fs.readdirSync(`${__dirname}/obo_node_json/`))
+
+describe('JSON to XML Test', () => {
+	for (let node of nodes) {
+		it(`Converts ${node}json to ${node}xml`, () => {
+			const json = fs.readFileSync(`${__dirname}/obo_node_json/${node}json`)
+			expect(jsonToXmlParser(json)).toMatchSnapshot()
```
iturgeon: All the snap shots for this test appear to be empty boiler plate xml.

packages/app/obojobo-document-json-parser/bin/json2xml.js
```diff
@@ -0,0 +1,11 @@
+const argv = process.argv.slice(2)
```
iturgeon: Both files in bin should be `chmod 775`

packages/app/obojobo-document-json-parser/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "dependencies": {
+    "obojobo-document-engine": "^4.0.1"
+  },
+  "devDependencies": {
+    "lint-staged": "^6.1.1"
+  },
+  "name": "json-to-xml-parser",
```
iturgeon: name needs to match the directory name as we'll publish this to npm. needs to be `obojobo-document-json-parser`

packages/app/obojobo-document-json-parser/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "version": "4.0.1",
+  "license": "AGPL-3.0-only",
+  "main": "",
+  "scripts": {
+    "test": "TZ='America/New_York' jest --verbose",
+    "test:ci": "TZ='America/New_York' CI=true jest --ci --coverage --coverageReporters text-summary cobertura",
+    "convertJsonToXml": "node ./bin/json2xml.js examples/test-object.json > test-object.xml",
```
iturgeon: Suggest adding a `bin` directive to make the xml converter scripts easier to use. See obojobo-express/package.json

vutoan1245:
> * The CLI ideally should be able to accept a path to any json file - right now it needs to be located at the top level of the package. So for example it would be best if node some/dir/json2xml.js ../../some/other/dir/file.json could work
> 
@zachberry You can do that now. But the path has to be the exact path or relative to the json2xml.js

vutoan1245:
@iturgeon Please review updated files

packages/app/obojobo-document-json-parser/__tests__/__snapshots__/jsonToXml.test.js.snap
```diff
@@ -0,0 +1,468 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`JSON to XML Test Converts assessment.json to assessment.xml 1`] = `
+"<?xml version=\\"1.0\\" encoding=\\"utf-8\\"?>
+<ObojoboDraftDoc>
+	<Assessment title=\\"Assessment\\" review=\\"always\\" rubric=\\"[object Object]\\" attempts=\\"5\\" id=\\"my-assessment\\">
```
zachberry: There are a couple of `[object Object]` attributes I saw - Like `rubric` and `actions` on ActionButton.

packages/app/obojobo-document-json-parser/__tests__/obo_node_json/figure.json
```diff
@@ -0,0 +1,21 @@
...
+		"url": null,
+		"size": "small",
+		"width": null,
+		"height": null,
+		"textGroup": [
+			{
+				"data": null,
+				"text": {
+					"value": "",
+					"styleList": []
```
zachberry: Would prefer a test case with text here

packages/app/obojobo-document-json-parser/src/actionButtonNodeParser/actionButtonNodeParser.js
```diff
@@ -0,0 +1,72 @@
+const textGroupParser = require('../textGroupParser')
+const xmlEncode = require('../xmlEncode')
+
+const actionButtonNodeParser = node => {
+    const id = node.id ? ` id="${node.id}"` : ''
+
+    let attrs = ''
+    for (const attr in node.content) {
+        if ([attr] == "triggers" || [attr] == "textGroup" || [attr] == "actions") continue;
```
iturgeon: in many of these parsers, the code `[attr] == "triggers"` should be simplified to `attr == "triggers"

iturgeon: Also, I'd like to use a blacklist for the things that must be removed:

```
   const blackListedAttributes = ['triggers', 'scoreActions', 'rubric']

    let attrs = ''
    for (const attr in node.content) {
        if(blackListedAttributes.includes(attr)) continue;
        attrs += ` ${[attr]}="${xmlEncode(node.content[attr])}"`
    }
```

But that leads to the idea of extracting this into a reusable util common to all the parsers so that we would end up with something like `const attrs = processAttrs(node.content, blacklist)`

iturgeon:
I just noticed file naming conventions here.  Obojobo uses kebab-case naming convention for file names.  There's a bunch of camel-case used in this PR.

PR #864, opened 2019-06-19
General Model Returning and promise cleanup

### LINKED ISSUE #863: Media.storeImageInDb returns more than needed when inserting binary data
Currently the insert binary data query returns *, but only uses ID.

Limiting it to `.id` automatically cleans up some of the code.

iturgeon:
* Multiple places where an insert was requesting * be returned from postgres when only id is used
* Some simplification of query and transaction promise structure
* made a pass to change `.then(result => {...})` to `.then(insertDraftResult => {...})` to clarify some confusing code
* simplified a couple places where variable renaming could simplify the source
* added comment to insert_event to explicitly say why we're not waiting for a result
* updated affected tests and snapshots
* `express_lti_launch.js`'s private `storeLtiLaunch` return changed from `{id: 3}` to `3` (and uses updated)

fixes #863


iturgeon:
Verified again after rebasing - I'm not getting any image errors now using the xml editor or the visual editor

PR #866, opened 2019-06-20
Fix non-failing errors showing up in test output

iturgeon:
The tests are printing a lot of errors that are being ignored.  To make sure we aren't ignoring something important, this is an attempt to clean all of those up.

Some Examples from the output

```
Warning: A component is changing a controlled input of type checkbox to be uncontrolled
```

```
Warning: Received the string `true` for the boolean attribute `allowFullScreen`. Although this works, it will not work as expected if you pass the string "false". Did you mean allowFullScreen={true}?
```

```
(node:8990) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'message' of undefined
```


iturgeon:
I cleaned up this pr a bit.  Rebased on dev/6 and smashed together the updates needed.  I omitted the prettier pass that needs to be run to keep things a little clearer

Issue #867, opened 2019-06-24
Creating two empty pages in the visual editor causes duplicate ID issue


zachberry:
To replicate:

1. Create a new Visual Editor tutorial
2. Create a new page, call it "a" (Don't edit anything in the page)
3. Create another new page, call it "b" (Don't edit anything in the page)

Result: Draft isn't saved because the children of the pages end up having the same IDs

PR #872, opened 2019-06-27
Empty page duplicated id

vutoan1245:
Reason for the problem: when a user adds two empty `Page`, Obojobo generates the same id for its `children`.
Solution: create a new function that generates unique id for `children`.

Fixed the same issue with two empty `Assessment`.

Resolve issue #867


iturgeon:
Verified, I have the same issue - I'll look into fixing

iturgeon:
Ugh.  There's a problem with how the templates are being used - page objects are bleeding into one another since they're actually sharing references via the original template.

Issue #491, opened 2018-10-01
Dialog needs a min-width attribute


iturgeon:
<a href="https://github.com/zachberry"><img src="https://avatars2.githubusercontent.com/u/73479?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zachberry](https://github.com/zachberry)**
_Monday May 07, 2018 at 20:19 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Document-Engine/issues/195_

----

The dialog that shows when there's an incomplete assessment can stack buttons on top of one another if the browser window is narrow enough. I believe the dialog can be given a width property - we probably also need to be able to specify a minWidth property. Then, give this dialog a minWidth such that it doesn't stack buttons on top of one another.

PR #874, opened 2019-06-27
Issue/491 incomplete assessment dialog render incorrectly on small screen

vutoan1245:
Add min-width attribute for incomplete assessment dialog so that buttons do not stack on each other.

Resolve issue #491


vutoan1245:
Wrong branch

PR #879, opened 2019-07-01
Upgrade to React 16.8.6

vutoan1245:
Upgrade to React 16.8.6

Resolve issue #879


package.json
```diff
@@ -71,5 +71,9 @@
 	"engines": {
 		"yarn": "^1.0",
 		"node": "^10.4.1"
+	},
+	"dependencies": {
```
iturgeon: I don't think these should be added at the top level project.json, will it work without it?

packages/util/obojobo-lib-utils/package.json
```diff
@@ -47,8 +47,8 @@
 		"jest": "24.1.0",
 		"lint-staged": "^6.1.1",
 		"prettier": "1.11.1",
-		"react-dom": "16.7.0",
-		"react-test-renderer": "16.7.0",
+		"react-dom": "16.8.6",
```
iturgeon: I believe we shouldn't have react-dom here under devDeps since its already under regular deps.

iturgeon:
Oh, wanted to add, this should have been pushed to your fork.  I changed some permission settings given some new features Github has, we'll see if that'll do what we want.

iturgeon:
Forgot to mention, I quickly tested the viewer and editor, the jump to a newer React seems to work just fine.

vutoan1245:
@iturgeon Sorry I was testing Github Desktop, not sure why it pushes directly to the repo.

vutoan1245:
I will close this PR and make a new one.

PR #889, opened 2019-07-09
Dev/7 carnelian

iturgeon:
Placeholder for dev/7.  Based off the current version of dev/6.


packages/app/obojobo-express/express_current_visit.js
```diff
@@ -8,6 +8,12 @@ const getCurrentVisitFromRequest = req => {
 		return Promise.resolve()
 	}
 
+	// In certain cases, events are sent via `navigator.sendBeacon`.
+	// Request's headers cannot be set. Therefore, `req.body` is still in JSON fotmat
+	try {
+		req.body = JSON.parse(req.body)
+	} catch (e) {} // eslint-disable-line no-empty
```
iturgeon: This will run JSON.parse on any request getting the current visit, check if there's a way to limit it to when sendBeacon is used.

packages/app/obojobo-express/config.js
```diff
@@ -85,12 +80,12 @@ const getConfigFileData = (configFile, env) => {
...
 
-		return Object.assign({}, defaultObject, envObject) // combine with default if it exists
+		// combine with default if it exists
+		return deepFreeze({ ...defaultObject, ...envObject })
 	} catch (error) {
 		logger.error(`Error loading config file: ${configFile}`)
 		logger.error(error.toString())
-
-		return {}
+		return deepFreeze({})
```
iturgeon: These deepFreeze calls may be redundant since it's called again on the whole config object afterwards.  Although - if we expose `getConfigFileData` it would still be useful?

Issue #749, opened 2019-03-11
IFrame hover labels stay stuck on when clicked


zachberry:
A side-effect of the accessibility pass, but the iframe labels stay on for the controls when clicked, requiring you to click off to remove them. Perhaps need to maintain this behavior when "clicked" with a keyboard press but not when clicked with a mouse.

PR #892, opened 2019-07-11
IFrame hover labels stay stuck on when clicked

vutoan1245:
Update button point state in IFrame so that it doesn't stay stuck when clicking

Resolve issue #749


zachberry:
I tweaked this a little, I wanted to make sure the hovered-over label was always on top. @vutoan1245 if you don't make taking a look.

PR #895, opened 2019-07-12
adds ability to launch an https server when not using a proxy

### LINKED ISSUE #894: Need ability to start the express server with ssl certs

iturgeon:
Adds a new optional param to startServer needed to pass the ssl certs to node.  The new option gets passed directly to `http(s).createServer` as is required to pass key and cert info.  

If the option object contains the `cert` key, obojobo will use node's `https` instead of `http` in order to handle those requests.

If you want to run both http and https servers, you'll need to run startServer twice, or start your own (useful for a super simple http -> https redirect)

```javascript
const startServer = require('obojobo-express/http_server')

// before this change
startServer(app, logger, 3000) 

// after this change
startServer(app, logger, 443, {cert: '...', key: '...'}) 
```
Fixes #894

### DIFF
--- a/packages/app/obojobo-express/http_server.js
+++ b/packages/app/obojobo-express/http_server.js
@@ -1,6 +1,4 @@
 /* eslint-disable */
-var http = require('http')
-
 /**
  * This code is derived from bin/www used in Express 4
  * We extracted the code from that file to make it 1: testable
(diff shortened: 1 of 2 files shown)


packages/app/obojobo-express/http_server.js
```diff
@@ -25,12 +23,14 @@ const normalizePort = val => {
...
 
-const startServer = (app, logger, port = '3000') => {
+const startServer = (app, logger, port = '3000', options = {}) => {
 	console.log('Note: Logging and config options can be set using environment variables.')
+	console.log(`DEBUG is set to: ${process.env.DEBUG}`)
 
 	port = normalizePort(port)
 	app.set('port', port)
-	const server = http.createServer(app)
+	const http_or_https = options.cert ? require('https') : require('http')
```
zachberry: Should we throw an error here if they pass in the 'cert' option but omit the 'key' option, or just let it ride?

iturgeon: I'm fine with relying on `https` to complain about missing something.

Issue #652, opened 2019-01-03
Heartbeat visit check to prevent multiple windows / check login


zachberry:
Users can have two Obo modules open. Need to implement some sort of heartbeat check and set other open windows into a locked state. This could also be used to ensure the user is still logged in (or the LTI hasn't expired), and lock the module when this has passed. Optionally show a dialog that the lock-out will happen soon.

iturgeon:
Just looking, we may able to use https://www.npmjs.com/package/sysend

PR #920, opened 2019-08-04
Multiple viewer window check

iturgeon:
solves issu #652 

* adds 'NoButtonModal' because all the other modals have close buttons
* moves visitId, draftId resolution out of viewer-app, up a level
* broadcasts to all other open tab when another tab is opened - older tabs are locked down.
* adds a heartbeat to verify with the server that the current visit is still valid, user is still logged in, and that the module hasn't been opened elsewhere
* launching a new visit of a module invalidates previous sessions (blocking other browsers, devices)


zachberry:
Discussed posting a new event when the duplicate window dialog is shown. There's an issue that the window that will post that event is an old visit that is now inactive, and therefore posting the event fails. None of the solutions seemed satisfactory and we decided not to go forward with this event. That discussion brought about talk of an event when a visit is deactivated which lead to #927

zachberry:
I hijacked this PR to include some missing linter and prettier updates

Issue #237, opened 2018-09-07
File manager for media uploads


iturgeon:
<a href="https://github.com/zachberry"><img src="https://avatars2.githubusercontent.com/u/73479?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zachberry](https://github.com/zachberry)**
_Monday Jun 18, 2018 at 14:20 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Next/issues/239_

----

zachberry:
To clarify - This would allow you to manage files you've uploaded

iturgeon:
For a starting point, here's how materia handles uploading files, I think the interface should mimic the workflow at least.

<img width="837" alt="Screen Shot 2019-08-06 at 11 40 45 AM" src="https://user-images.githubusercontent.com/73480/62554494-3699a480-b83f-11e9-9294-580d26f18695.png">

zachberry:
Some mockups

![image](https://user-images.githubusercontent.com/73479/62799325-87630480-baae-11e9-8564-3ccc9ff60df3.png)

![image](https://user-images.githubusercontent.com/73479/62799334-8af68b80-baae-11e9-9c77-213fd2614335.png)

PR #923, opened 2019-08-06
File upload manager

vutoan1245:
Add Image Upload Manager for Figure

Resolve issue #237


packages/obonode/obojobo-chunks-figure/image-properties-modal.js
```diff
@@ -123,99 +116,104 @@ class ImageProperties extends React.Component {
...
+							type="text"
+							id="obojobo-draft--chunks--figure--alt"
+							value={this.state.alt || ''}
+							onChange={this.handleAltTextChange.bind(this)}
+							size="50"
+							placeholder="Describe the Image"
+						/>
+
+						<label htmlFor="obojobo-draft--chunks--figure--size">Size:</label>
+						<div id="obojobo-draft--chunks--figure--size">
```
qwertynerd97: The label is not associated properly with the radio button group, so screenreader users cannot tell that size is what the radio buttons are describing.  This might be fixed by adding aria-role="radiogroup" to the div.

packages/obonode/obojobo-chunks-figure/choose-image-modal.js
```diff
@@ -0,0 +1,145 @@
...
+								id={media.id}
+								src={`/api/media/${media.id}/small`}
+								onClick={() => this.props.onCloseChooseImageModal(media.id)}
+								onKeyPress={event =>
+									this.onHandleKeyPress(event, () => this.props.onCloseChooseImageModal(media.id))
+								}
+							/>
+						))}
+
+						{this.state.isFetching ? <h5>Loading...</h5> : null}
```
qwertynerd97: Screen readers read out the "Loading..." regardless of whether or not the images have already loaded.

packages/obonode/obojobo-chunks-figure/image-properties-modal.js
```diff
@@ -123,99 +116,104 @@ class ImageProperties extends React.Component {
...
+										<span>px × </span>
+										<input
+											id="obojobo-draft--chunks--figure--custom-height"
+											name="custom-height"
+											min="1"
+											max="2000"
+											step="1"
+											type="number"
+											placeholder="Height"
+											aria-label="Height"
```
qwertynerd97: Sighted users can see that the unit is pixels for these fields, but this is not clear when using a screen reader.  Maybe change the aria-label to Height/Width in pixels

packages/obonode/obojobo-chunks-figure/choose-image-modal.js
vutoan1245: I added `aria-live="polite"`. User will be notified when the UI is updated

zachberry: Testing this I always ended up hearing "Loading...". I tried using hidden, aria-busy and role="alert" since I was curious but I also always ended up hearing 'Loading'. I'm not sure the solution to this one.

packages/obonode/obojobo-chunks-figure/image-properties-modal.js
```diff
@@ -123,99 +116,108 @@ class ImageProperties extends React.Component {
...
+							value={this.state.alt || ''}
+							onChange={this.handleAltTextChange.bind(this)}
+							size="50"
+							placeholder="Describe the Image"
+						/>
+
+						<label htmlFor="obojobo-draft--chunks--figure--size">Size:</label>
+						<div
+							id="obojobo-draft--chunks--figure--size"
+							aria-role="radiogroup"
```
zachberry: `aria-role` isn't valid:
```suggestion
							role="radiogroup"
```

packages/obonode/obojobo-chunks-figure/choose-image-modal.js
zachberry: @qwertynerd97 Just checking if you have any ideas?

qwertynerd97: The problem is that the content of the dialog is loaded into the read buffer with Loading..., and it reads that out, even though the content area has changed (also i think ChromeVox may have some issues with aria-live?) .  For the aria-live to work, it needs to be on an element that is loaded with the DOM - moving it up to the div might work since it's loaded with the modal, but I'm not certian.

iturgeon:
Note while testing: If I change an image then open the `...` dialog and hit save, it reverts to the previous image.

iturgeon:
I made an attempt at the aria tags - It'll now tell you it's loading as it's reading through the modal.  After it reads the selected ok button, It'll read out "11 Recently uploaded images loaded" or "11 Recently uploaded images loaded with more" if there are more to load.  When load more is pressed, this label is re-read with the updated information.

packages/app/obojobo-express/server/routes/api/media.js
```diff
@@ -59,4 +59,21 @@ router
 			.catch(next)
 	})
 
+// Get list of media a thumbnail
+// mounted as /api/media/many
+router
+	.route('/many/')
+	.get([requireCurrentUser])
+	.get((req, res, next) => {
+		const start = parseInt(req.query.start, 10)
```
iturgeon: We need to implement some default/min/max values here and make sure start and count are positive ints.

We could use `express-validator` to make sure start and count are sent and are

packages/app/obojobo-express/server/models/media.js
```diff
@@ -210,6 +210,45 @@ class Media {
...
+	static fetchManyById(userId, start, count) {
+		if (!userId || !Number.isInteger(start) || !Number.isInteger(count)) {
+			throw new Error('Invalid argument.')
+		}
+
+		return db
+			.manyOrNone(
+				`
+					SELECT
+						id as "id",
```
iturgeon: I don't think `id as "id"` is necessary here

iturgeon:
this also fixes #1140

PR #913, opened 2019-07-26
Added size input to equation editor

AnthonyRodriguez726:
Added an input field on the equation editor that allows users to edit the font-size of their equations.

PR #924, opened 2019-08-06
Issue/901 size property

### LINKED ISSUE #901: Need to add 'size' property to the edit Math Equation dialog in the visual editor
See:

image

The 'size' attribute is missing. Should be a number field (default value of 1) that could step in 0.1 increments. Minimum should be set to a value > 0 such that any smaller would render the equation unreadable.

AnthonyRodriguez726:
Added live equation preview in the editor. This is an addition to #913 and resolves #901


packages/obonode/obojobo-chunks-math-equation/math-equation-properties-modal.js
```diff
@@ -112,6 +159,16 @@ class MathEquationProperties extends React.Component {
 					</div>
 					<span>{this.state.error}</span>
 				</div>
+				<div id="katexContainer" style={{ display: this.latexCheck(this.state.content.latex) }}>
+					<b id="previewTitle">Preview</b>
```
zachberry: For css classes/ids [use `kebab-case` instead of `camelCase`](https://github.com/ucfopen/Obojobo/wiki/Developer-Code-Style-Guide#class-naming-conventions)

zachberry: Same with 'katexContainer', 'katexPreview', and any others I might have missed

packages/obonode/obojobo-chunks-math-equation/math-equation-properties-modal.js
```diff
@@ -112,6 +159,16 @@ class MathEquationProperties extends React.Component {
 					</div>
 					<span>{this.state.error}</span>
 				</div>
+				<div id="katexContainer" style={{ display: this.latexCheck(this.state.content.latex) }}>
```
zachberry: I think we should always show element this since it causes the dialog to suddenly expand as soon as you type something into the first input.

However, I did want to give some feedback on this (as if we were gonna keep it) for instructional purposes. I feel that latexCheck isn't a descriptive name because it doesn't tell you that it's going to return a string of 'none' or 'block'. Instead, I think latexCheck should simply return a boolean, and you can call it something like `isLatexValid`. So if we were gonna keep this then this line would become

`style={{ display: this.isLatexValid(this.state.content.latex) ? 'block' : 'none'}}`

That way we have a simple method we can test, isLatexValid, which is descriptive, and the specific css-relevant values of 'block' and 'none' are kept inside the CSS we're writing.

packages/obonode/obojobo-chunks-math-equation/math-equation-properties-modal.js
```diff
@@ -112,6 +159,16 @@ class MathEquationProperties extends React.Component {
...
+					<b id="previewTitle">Preview</b>
+					<div
+						style={{ fontSize: this.state.content.size + 'em' }}
+						aria-hidden
+						id="example-katex"
+						className={'katexPreview'}
+						dangerouslySetInnerHTML={{ __html: this.equationGenerator() }}
```
zachberry: I prefer to use a `get` named method in this case, like `getEquationHTML`, it's more action oriented where as equationGenerator is more passive. I think `generateEquation` would work too but that still doesn't tell you that it is returning HTML, and I like standardizing on using `get` for any function that is getting something for you.

zachberry: https://github.com/ucfopen/Obojobo/wiki/Developer-Code-Style-Guide#function-naming-conventions

packages/obonode/obojobo-chunks-math-equation/math-equation-properties-modal.scss
```diff
@@ -32,3 +36,18 @@
 		}
 	}
 }
+
+#katexContainer {
```
zachberry: * Always prefer to use classes instead of ids in any component.
* This style should be nested inside the `.math-equation-properties` rule. As is this rule is now going to globally style anything on the page with an id of 'katexContainer'. While it's unlikely you still can't be 100% certain that another developer created a content type in Obojobo that is also using an element with an id of `katexContainer`.

PR #934, opened 2019-08-13
adds check for learner role in sendHighestAssessmentScore

### LINKED ISSUE #808: In Canvas lis_result_sourcedid is not coming through for Teachers or Test Student
Prior to v3.4.0 (Amethyst) preview mode was always shown for Teachers but this was changed. It seems that Canvas is not sending the lis_result_sourcedid through in that case, resulting in LTI replaceResult failures.

UI/UX wise this means that a Faculty member will go to submit an assessment score and see the error message that it wasn't sent, and likely could assume it's not working (however it should be working for their students).

I'm not sure if we should go back to the pre-v3.4.0 behavior or implement a special case that maybe reverts to preview mode if the member has the proper role but no lis_result_sourcedid is found.

iturgeon:
WIP - only half done.  Looking for some feedback.

Right now, this only adds a new error and error message on the server side when attempting to send the score via LTI.  The client side would need to handle this.

Not entirely sure what we want to do UI wise for this case - should a user w/o Learner role even be able to start a non-preview visit?

Fixes #808

### DIFF
--- a/packages/app/obojobo-express/__tests__/lti.test.js
+++ b/packages/app/obojobo-express/__tests__/lti.test.js
@@ -72,7 +72,8 @@ const mockSendAssessScoreDBCalls = (
 	ltiHasOutcome, // Use 'missing' to indicate no launch, 'error' to throw an error
 	ltiKey = 'testkey',
 	insertLTIAssessmentScoreSucceeds = true,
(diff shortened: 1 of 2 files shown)


iturgeon:
Closing for now

Issue #936, opened 2019-08-13
Updates to Preview Mode


zachberry:
Discussed with @iturgeon a plan for upgrades to Preview Mode, which came out of a discussion about #808 

This would be a multi-phase plan - I'm making this as a single issue but we can create additional issues out of this issues.

Phase 1:
* Revert preview mode back to pre 3.3 behavior so that any user that comes in with a Teacher-esq role views objects in preview mode. This solves #808 and the confusion that comes with that

Phase 2:
* Build in Google Docs-style permissions. Instructors/authors could make modules public or private, and add people who can preview. That makes preview mode more restricted for better security, and solves issues with an instructor in course A being able to preview modules in course B.
* Build a preview landing page. This page can explain preview mode and provide buttons to provide access if they don't have access.

Phase 3:

* Sample mode - Being able to preview a portion of a module. This is used when "shopping" for a module, which prevents issues from instructors who have access to the future Public Library being able to view too much of the content.

PR #938, opened 2019-08-14
launches all non-learners into preview mode

iturgeon:
Phase 1 of #936 (also a mention for #808 )

* adds `canCompleteAssignment` to user permissions
* uses `canCompleteAssignment` to determine if a user can view and assignment or should be preview it
* changes `req.setCurrentUser` to store a user in req.currentUser since the user is already loaded
* changes `req.getCurrentUser` to store the GuestUser in req.currentUser to be consistent with how it works for logged in users
* tests
* changes `canViewDrafts` to `canPreviewDrafts` in perms
* changes `requireCanViewDrafts` to `requireCanPreviewDrafts` in validators


zachberry:
Fixes #808

packages/app/obojobo-express/config/permission_groups.json
```diff
@@ -4,7 +4,8 @@
 		"canEditDrafts": ["Instructor", "Administrator", "urn:lti:instrole:ims/lis/Administrator", "urn:lti:role:ims/lis/TeachingAssistant"],
 		"canDeleteDrafts": ["Instructor", "Administrator", "urn:lti:instrole:ims/lis/Administrator", "urn:lti:role:ims/lis/TeachingAssistant"],
 		"canCreateDrafts": ["Instructor", "Administrator", "urn:lti:instrole:ims/lis/Administrator", "urn:lti:role:ims/lis/TeachingAssistant"],
-		"canViewDrafts": ["Instructor", "Administrator", "urn:lti:instrole:ims/lis/Administrator", "urn:lti:role:ims/lis/TeachingAssistant"]
+		"canViewDrafts": ["Instructor", "Administrator", "urn:lti:instrole:ims/lis/Administrator", "urn:lti:role:ims/lis/TeachingAssistant"],
+		"canCompleteAssignment": ["Learner", "urn:lti:role:ims/lis/Learner"]
```
zachberry: The name for this trips me up a little, because it's really 'canViewTheActualModuleNotPreviewMode` (but of course we can't call it that). Because if this is false then they are forced into preview mode. And technically they _can_ complete the assignment, it's just that we probably can't send a score to any gradebook via LTI.

Also, and this isn't in regards to this PR, there is a 'canViewDrafts' perm which doesn't include Learner, which seems confusing since Learners can view drafts.

So one idea: Add Learner to canViewDrafts and re-evaulate how that perm is being used. Then perhaps add a `canOnlyPreviewDrafts` which would contain everyone that canViewDrafts but exclude Learner.

Not sure I like it. Tossing out some other name ideas: `canSubmitAssessment`, `canSubmitAttempt`, `canSendLTIScore`.

packages/app/obojobo-express/__tests__/routes/viewer.test.js
```diff
@@ -145,6 +145,29 @@ describe('viewer route', () => {
...
+		expect.assertions(3)
+		mockCurrentUser.canCompleteAssignment = false
+		const uuid = validUUID()
+
+		VisitModel.createVisit.mockResolvedValueOnce({
+			visitId: 'mocked-visit-id',
+			deactivatedVisitId: 'mocked-deactivated-visit-id'
+		})
+
+		// mockCurrentUser = {id: 44}
```
zachberry:
```suggestion
```

packages/app/obojobo-express/express_current_user.js
```diff
@@ -5,6 +5,7 @@ const logger = oboRequire('logger')
 const setCurrentUser = (req, user) => {
 	if (!(user instanceof User)) throw new Error('Invalid User for Current user')
 	req.session.currentUserId = user.id
```
zachberry: Do we need to set this? It seems to only be used in this file.

iturgeon: req.session.currentUserId?  yea, this is where the server links the current session to the user (req.session is written to the db)

packages/app/obojobo-express/config/permission_groups.json
iturgeon: we landed on `canViewAsStudent`

iturgeon: I renamed canViewDrafts to canPreviewDrafts.

It was pretty much used the same way, so I think it works.

packages/app/obojobo-express/routes/api/drafts.js
```diff
@@ -189,7 +189,7 @@ router
 // mounted as /api/drafts
 router
 	.route('/')
-	.get(requireCanViewDrafts)
+	.get(requireCanPreviewDrafts)
```
zachberry: I like the new `canPreviewDrafts`, this is the only case where it feels a bit weird to use. In other words, this is "If user X can preview drafts then return a list of drafts that user X has".

I'm thinking do we even need this check here? Could we always let users get a list of their drafts? Students (as the system is) would simply get nothing returned as they don't have drafts.

PR #940, opened 2019-08-14
Lti code cleanup

iturgeon:
* changes to prevent mock user from conflictng with guest user
* lti launch code now adopting res.currentUser and res.currentDocument
* Updated text for user validation to include msg that it may be permissions OR missing user
* simplified a complex double ternary checking for which return url to use


packages/app/obojobo-express/express_lti_launch.js
```diff
@@ -125,22 +125,13 @@ exports.assignment = (req, res, next) => {
...
-
 	return Promise.resolve(req.lti)
 		.then(lti => userFromLaunch(req, lti.body))
-		.then(launchUser => {
-			currentUser = launchUser
-			return req.requireCurrentDocument()
-		})
-		.then(draftDocument => {
-			currentDocument = draftDocument
+		.then(() => req.requireCurrentDocument())
```
zachberry: Are we able to do this?

```suggestion
		.then(req.requireCurrentDocument)
```

packages/app/obojobo-express/routes/lti.js
```diff
@@ -51,11 +51,8 @@ const showModuleSelector = (req, res) => {
 		let returnUrl = null
 		let isAssignment = false
 		if (req.lti && req.lti.body) {
-			returnUrl = req.lti.body.content_item_return_url ? req.lti.body.content_item_return_url : null
-
-			returnUrl = req.lti.body.ext_content_return_url
-				? req.lti.body.ext_content_return_url
-				: returnUrl
+			returnUrl =
+				req.lti.body.ext_content_return_url || req.lti.body.content_item_return_url || null
```
zachberry: Much better

packages/app/obojobo-express/express_lti_launch.js
iturgeon: I struggle with this - Here's my thoughts on a conventsion: if the function being called doesn't use the return from the previous promise, then it should isolate the method like in the PR.

zachberry: I can get behind that

PR #941, opened 2019-08-14
Adds student launch options to /dev lti links

iturgeon:
Adds launch as student options to /dev links


zachberry:
@qwertynerd97 Agree, created #969 to make sure we do this

PR #963, opened 2019-09-06
Adds isolation to new nodes in visual editor Level 2

### LINKED ISSUE #961: all YouTube ids are synced in the Visual Editor
If you create 2 youtube videos in the visual editor with different Ids, the last one you entered changes both.

Note: if you reload the editor after saving a page, and create a new video - then that one will appear to work, but any additional videos will be linked to the one you created.

I suspect this is due to the way new objects are created when you hit the + button.  - They probably return a reference to the original object causing them all to be linked together.

iturgeon:
This is a more complete/opinionated solution to issue #961 than this PR #962 

@zachberry, I offer this as a second PR so we can either merge this one into master, or merge 962 in for simplicity & speed, then merge this into the next dev branch?

Fixes #961 

* changes `insertJSON` -> `templateObject` in each chunk's call to `registerModel`
* adds a method to the model that will clone a templateObject called `cloneBlankNode`
* adds a test for `cloneBlankNode`
* Registry.registerModel looks like it was creating some option params in a strange order & they weren't being used - so I removed them and re-ordered
* removes .json.emptyNode from each node's exports to reduce unintentional pass by ref ofthe  templateObject (and this is not currently used)

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/common/registry.js
+++ b/packages/app/obojobo-document-engine/src/scripts/common/registry.js
@@ -4,6 +4,8 @@ let registeredToolbarItems
 let toolbarItems
 let variableHandlers
 
+const noop = () => {}
+
 class _Registry {
 	init() {
 		items = new Map()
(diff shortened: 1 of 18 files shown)


packages/app/obojobo-document-engine/src/scripts/common/registry.js
```diff
@@ -40,34 +40,29 @@ class _Registry {
...
-				// pull out to editor registry eventually
-				name: '',
-				icon: null,
-				isInsertable: false,
-				slateToObo: null,
-				oboToSlate: null,
-				plugins: null,
-				supportsChildren: false
+				templateObject: '',
+				init() {}
```
zachberry: The removed properties seem useful to ensure that a node has all of these in their opts config. I'm sure I'm missing something but why were these removed?

iturgeon: https://github.com/ucfopen/Obojobo/blob/35d5eee807ed08cf36ac02a808876bcf2ef6bf18/packages/app/obojobo-document-engine/src/scripts/common/registry.js#L49

Note that `opts` is passed into `items.set`, THEN `opts` is re-assigned a new value (where it gets these default properties).  At this point opts will be a different object than the one sent to `items`.

I don' t think this is the intended purpose AND it proves those properties were never used.

zachberry: Ahh, makes sense to me now

PR #978, opened 2019-10-03
Implement jest-runner-eslint

### LINKED ISSUE #743: look into adding jest-eslint-runner for quality of life
This could be a really nice to have item https://github.com/jest-community/jest-runner-eslint

iturgeon:
Adds jest-runner-eslint to the mono-repo. 

Check it out to see what you thing:

* there will be a few files that need to be cleaned up to pass
* we could remove eslint:js from the precommit tasks
* jest eslint will only run when running jest at the top level

fixes #743


packages/app/obojobo-document-xml-parser/src/extension-transform.js
```diff
@@ -28,29 +26,26 @@ let toHTML = el => {
 	)
 }
 
-let extensionTransform = node => {
+const extensionTransform = node => {
 	// console.log('node', node)
```
iturgeon:
```suggestion
```

iturgeon:
NOTE: I changed the precommit tests to bail on the first failure instead of finding them all.  Makes sense to me for now.

iturgeon:
Ok I caught it up to dev/8 and found an error being logged in the tests

PR #980, opened 2019-10-08
adds loading=lazy to img and iframe tags

iturgeon:
To speed up loading - add in a newly supported loading="lazy" attribute on img and iframe tags.

fixes #980


zachberry:
@iturgeon I implemented a few tweaks, please review and let me know what you think:

1. Gray background on images so on slow connections students know that an image will appear (previously this was white and wasn't clear what was happening on the page)
2. Doesn't lazy load in the editor (Open to discussion if this is better or worse)

PR #984, opened 2019-10-08
General bug fixes for dev7 after responsive code

iturgeon:
* Reduces number of events sent to the server
* Centralizes debounce code
* Adds tests to nav for use cases of multiple clicks
* fixes bug that kept nav:toggle button actions from having an effect on mobile when the nav was open
* nav now only opens when increasing from mobile width to non-mobile width
* fix code to use correct font
* changes latex labes to use font used for figure labels
* removes commented out code
* fixes math equation left/right/center alignment after katex version update

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/viewer/components/nav.js
+++ b/packages/app/obojobo-document-engine/src/scripts/viewer/components/nav.js
@@ -11,6 +11,7 @@ const { StyleableText, StyleableTextComponent } = Common.text
 const { isOrNot } = Common.util
 const { Dispatcher } = Common.flux
 const TIMER_MOBILE_NAV_COLLAPSE = 900
+const IS_MOBILE_MEDIA_QUERY = '(max-width: 480px)'
 
 const getLabelTextFromLabel = label => {
 	if (!label) return ''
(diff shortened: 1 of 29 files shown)


packages/app/obojobo-document-engine/src/scripts/viewer/components/nav.js
```diff
@@ -11,6 +11,7 @@ const { StyleableText, StyleableTextComponent } = Common.text
 const { isOrNot } = Common.util
 const { Dispatcher } = Common.flux
 const TIMER_MOBILE_NAV_COLLAPSE = 900
+const IS_MOBILE_MEDIA_QUERY = '(max-width: 480px)'
```
zachberry:
```suggestion
const MOBILE_MEDIA_QUERY = '(max-width: 480px)'
```

packages/app/obojobo-document-engine/src/scripts/viewer/components/nav.js
```diff
@@ -21,48 +22,52 @@ export default class Nav extends React.Component {
...
 		super(props)
 		this.selfRef = React.createRef()
-		this.onWindowClick = this.onWindowClick.bind(this)
 		this.hideOrShowOnResize = this.hideOrShowOnResize.bind(this)
+		this.closeNavOnMobile = this.closeNavOnMobile.bind(this)
 	}
 
 	isMobileSize() {
-		return window.matchMedia('(max-width: 480px)').matches
+		return window.matchMedia(IS_MOBILE_MEDIA_QUERY).matches
```
zachberry:
```suggestion
		return window.matchMedia(MOBILE_MEDIA_QUERY).matches
```

PR #1000, opened 2019-11-01
api/visit/start errors sooner when student visits an old draft version

### LINKED ISSUE #999: api/visit/start should bail sooner when student visits an old draft version
The check for students visiting an older draft bails a little late here: 
packages/app/obojobo-express/routes/api/visits.js:79-85

It could bail before getting the viewer state and props.

iturgeon:
When starting a draft visit, there's a check to see if
there's a newer version of the draft - which blocks a student from
viewing old versions of a module.

Previously the code would get the viewer state and props before bailing,
doing some unnecessary work.

fixes #999


packages/app/obojobo-express/routes/api/visits.js
```diff
@@ -61,28 +61,32 @@ router
...
 				throw 'Unable to start visit, visitId is no longer valid'
 			})
-			.then(() =>
-				Promise.all([
+			.then(() => {
+				// error so the student starts a new view w/ newer version
+				// this check doesn't happen in preview mode so authors
+				// can reload the page to see changes easier
+				if (
+					req.currentVisit.is_preview === false &&
```
zachberry:
```suggestion
					!req.currentVisit.is_preview &&
```

Issue #89, opened 2018-09-07
Test Issue: Save and read red alert status


iturgeon:
This a test issue to help get familiar with the workings of the project.

Need to add a button to the left-hand nav in the viewer that, when pressed, turns the background of the Navigation to red. Pressing it again would return the nav back to normal. Also need to log when the nav is set to the red alert state and returned back to the normal state.

If you haven't yet, fork the main repo. You'll be working off of your own branch on your own fork. Next, create a new branch to work on - `git checkout -b issue/89-red-alert-status` (off of your fork).

You'll also want to check out the documentation at https://github.com/ucfopen/Obojobo-Docs and also the project code style guide at https://github.com/ucfopen/Obojobo/wiki/Developer-Code-Style-Guide

Here would be the steps to implement this:

# In the `obojobo-document-engine` package:

* Add a new event handler to **NavStore** to listen to `nav:redAlert`. The payload of the event should include a boolean to indicate if red alert is on or off. This event handler should call `APIUtil.postEvent` to send a `nav:redAlert` event to the server.
* Modify the state of **NavStore** to include a new `redAlert` property holding a boolean if red alert is enabled or not.
* Add a new method to **NavUtil** called `setRedAlert` to fire the `nav:redAlert` event. The method should take a boolean to indicate if red alert should be enabled or not.
* Add a button to the nav component. Clicking this should call the **NavUtil.setRedAlert**  method.
* Add a query method to **NavUtil** - `isRedAlertEnabled` - that will return a boolean to indicate if red alert is enabled or not
* Add two new CSS classes in the nav component - `is-red-alert` and `is-not-red-alert`. The nav component should use the **NavUtil.isRedAlertEnabled** method to figure out which class to apply to the element.
* Add a `is-red-alert` style to the nav component scss file which turns the background of the nav red.

In addition, you'll need to ensure 100% test coverage. This is not an exhaustive list but here are a few example things you'll want to test:

* [ ] - In viewer-app.js test that clicking the button toggles between setting the `is-red-alert` and `is-not-red-alert` classes on the appropriate element.
* [ ] - Add a NavUtil test ensuring that **NavUtil.setRedAlert** calls the Dispatcher.trigger method with the correct parameters and that **NavUtil.isRedAlertEnabled** returns expected values with given inputs.
  
# In the `obojobo-express` package:

[...]

PR #1006, opened 2019-11-20
Issue/89 red alert status

ctcuff:
A test issue that adds a button to the nav bar turning it red

### DIFF
--- a/packages/app/obojobo-express/migrations/20191108191111-red-alert-status.js
+++ b/packages/app/obojobo-express/migrations/20191108191111-red-alert-status.js
@@ -30,11 +30,11 @@ exports.up = function(db) {
 				type: 'UUID',
 				notNull: true
 			},
-			content_id: {
+			draft_content_id: {
 				type: 'UUID',
 				notNull: true
 			},
-			creation_time: {
+			created_at: {
 				type: 'timestamp WITH TIME ZONE',
 				notNull: true,
 				defaultValue: new String('now()')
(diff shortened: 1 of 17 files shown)


packages/app/obojobo-express/migrations/20191108191111-red-alert-status.js
```diff
@@ -0,0 +1,59 @@
...
+			},
+			draft_id: {
+				type: 'UUID',
+				notNull: true
+			},
+			content_id: {
+				type: 'UUID',
+				notNull: true
+			},
+			creation_time: {
```
iturgeon: just worth noting, we'd use `created_at` here to match all the other tables.

packages/app/obojobo-express/routes/api/visits.js
```diff
@@ -116,12 +117,24 @@ router
 				})
 			})
 			.then(() => {
+				return db.oneOrNone(
+					`
+						SELECT is_red_alert_enabled from red_alert_status
+						WHERE user_id = $[userId] AND draft_id = $[draftId]
```
iturgeon: Note: this will turn on red alert for:
* this module used in other courses
* all versions of this module (if the instructor edits the draft, the alert persists)

Think about what it would take to isolate one or both of those conditions.

Hints: check out `draft_content_id` and `resource_link_id`

iturgeon:
Secondary note: the text in this issue should reference the issue somewhere (like this #89)

ctcuff:
Also forgot to mention that this fixes #89

zachberry:
Closing since this is a training exercise

PR #1007, opened 2019-12-06
Issue/525 youtube time values

### LINKED ISSUE #525: Add time values to YouTube chunk
_Wednesday Jun 13, 2018 at 18:02 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Document-Engine/issues/229_

----

Would be nice to have:

* `time`: What time to start the video at
* `endTime`: Optional end time (not sure if YouTube supports this) - With both `time` and `endTime` specified then a section of a video can be embedded.

ctcuff:
Adds an option to start or end videos at a certain point. Ended up wrapping [YouTube's IFrame API](https://developers.google.com/youtube/iframe_api_reference) into its own component.

Resolves #525


zachberry:
@qwertynerd97 We did discuss the inputs, and settled on adding the feature now and polishing it later. However perhaps we could resolve some of the confusion by simply adding "(Seconds)" to the input labels.

ctcuff:
@qwertynerd97 @zachberry Typing time values like `mm:ss` would be more clear but I wasn't sure if that was something I should include in this PR or move to a different issue. It would just involve parsing some input. Youtube's IFrame API only takes seconds though so something like `2:40` or maybe `2m40s`would have to be interpreted as 160. As for the XML, that's probably something I could fix.

qwertynerd97:
Yeah, I think leaving it as seconds is fine for now, as long as we add some sort of clarifying notation to the modal

PR #1008, opened 2019-12-09
Inserting and Formatting

qwertynerd97:
Requires #989 as a pre-requisite

Adds capabilities for inserting any node type into the document in the middle of other nodes

Adds in formatting options such as switching between node types and toggling list styles.  Formatting can be changed using the Format menu or by using the dropdown menus in the toolbar.
Note: Toggling formats across several nodes is currently broken due to the way slate formats a selection.  This will be resolved when we upgrade to the latest version of Slate


zachberry:
Testing locally, attempted to merge in #989, so feedback based on that. I've only tested lists and the heading/code/text menu so far, and this is just a first quick pass but I wanted to get it written down:

* Lists don't seem to get styled the way I expect them. For example, if at a level 2 list item I change it from disc to square then I'd expect every item at level 2 to change to square. However it only seems to change the item the cursor is on.
* The changes to list only seem to effect the editor. The viewer simply uses the defaults.
* Changing the text type menu to 'code' converts your selection to a 'code' node, but I sorta expect that to change the text to monospace styled.
* The text type menu should close after something is selected. And if it is open, clicking away should close it. It does close sometimes when clicking off but when clicking in the content it stays open.

qwertynerd97:
All of the changes from this PR were upgraded to Slate 0.52 and rolled into #1159

PR #1020, opened 2019-12-18
[clean] removes fetch shim needed to support ie11

iturgeon:
We no longer support ie11, so the fetch shim to do so can be removed.


zachberry:
@iturgeon If we're removing this for IE we probably should remove the polyfills we have in polyfills.js - Object.assign, window.Set, Array.from, window.Promise, window.Symbol and Number.isFinite

Will put in new issue

PR #1034, opened 2019-12-20
[fix] dashboard filtering with upper and lower case

iturgeon:
dashboard filtering didn't match uppercase against lower case strings


iturgeon:
Canceling and changing the target to a branch @zachberry had built to solve the same problem

PR #1043, opened 2019-12-23
[fix] Adds promise provider to fix shared module perms

iturgeon:
* adds promise_provider to emit promise based events
* adds user to module perm check in draft model
* registers user to module perm check from repository
* overhauls api/draft/:draftid/full to use new perm checking

### DIFF
--- a/packages/app/obojobo-express/models/draft.js
+++ b/packages/app/obojobo-express/models/draft.js
@@ -72,6 +72,20 @@ class Draft {
 		return draftNode
 	}
 
+	// returns a boolean
+	static async userHasPermissionToDraft(userId, draftId) {
+		const result = await db.oneOrNone(
+			`SELECT user_id
+			FROM drafts
+			WHERE id = $[draftId]
+			AND user_id = $[userId]`,
+			{ userId, draftId }
+		)
+
+		// oneOrNone reutrns null when there are no results
+		return result !== null
+	}
+
 	static deleteByIdAndUser(id, userId) {
 		return db
 			.none(
--- a/packages/app/obojobo-express/util/promise_provider.js
+++ b/packages/app/obojobo-express/util/promise_provider.js
@@ -1,0 +1,45 @@
+// possible npm packages to use instead:
+// emittery
+// p-event
+const logger = oboRequire('logger')
+const promiseProviders = {}
+
+// Register promise to an event
+// your provider mus be an async function or Promise
+const registerPromiseProvider = (eventName, provider) => {
+	if (!promiseProviders[eventName]) promiseProviders[eventName] = new Set()
+	const providersForType = promiseProviders[eventName]
+	providersForType.add(provider)
+}
+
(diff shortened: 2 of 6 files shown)


packages/app/obojobo-express/models/draft.js
```diff
@@ -72,6 +72,20 @@ class Draft {
 		return draftNode
 	}
 
+	// returns a boolean
+	static async userHasPermissionToDraft(userId, draftId) {
+		const result = await db.oneOrNone(
+			`SELECT user_id
+			FROM drafts
```
zachberry: Is this correct, or should it be pulling from the map_user_to_draft table?

packages/app/obojobo-express/util/promise_provider.js
```diff
@@ -0,0 +1,45 @@
...
+	if (!promiseProviders[eventName]) return []
+
+	const providersForType = promiseProviders[eventName]
+
+	const results = []
+	for (const provider of providersForType) {
+		try {
+			const result = await provider(eventName, providerArgs)
+			results.push(result)
+			if (result && returnOnFirstNonNullResponse) break
```
iturgeon:
```suggestion
			if (result !== null && returnOnFirstNonNullResponse) break
```

packages/app/obojobo-express/util/promise_provider.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// execute all promises for an eventName in order with the provided arugments
+// setting returnOnFirstNonNullResponse to true will return immediately
+// when a promise returns a non-null value, skipping the remaining items
+// always returns an array
+const callPromiseProviders = async (
+	eventName,
+	providerArgs,
+	returnOnFirstNonNullResponse = false
```
iturgeon: Notes to self: perhaps this could be `returnIfTrueFn` passing something like `value => value === true` would return as soon as a `true` is found, but be flexible enough to handle other options.

packages/app/obojobo-express/models/draft.js
iturgeon: Here - yes. Keep in mind this PR is intended to keep `obojobo-express` decoupled from other packages.

From the perspective of obojobo-express, the only owner is the creator and needs that to operate without repository. obojobo-repository implements the concept of multiple owners (db tables, queries, etc.). The promiseProvider allows these decoupled libraries to manage permissions.

When the following is executed:
```
const hasPermResults = await callPromiseProviders('userHasPermissionToDraft', args, true)
```

The promise provider will call each provider in series. First obojobo-express's then any others.  In our current setup, obojobo-repository has a listener that will check the `repository_map_user_to_draft` table.

iturgeon:
I think I'm going to close this as it's not super valuable right now.  The complication of dealing with multiple responders, cascading down each responder, and the fact that this is only really used in once case now.  

Currently thinking the preferred solution is to either escalate the sharing permissions from repository to express or make a new shared module specifically for shared permissions that both use.

Issue #1049, opened 2019-12-30
Question bank editor settings radio buttons not selecting correctly


iturgeon:
The state of the selected radio inputs for choosing if all questions should be shown or if a subset should be shown don't always indicate the current value correctly, especially when using nested question banks.

PR #1050, opened 2019-12-30
[fix] editor question bank settings radio buttons

iturgeon:
* adds question bank id to editor setting radio button input name

The radio buttons for question bank's settings in the editor
all shared the same name. Selecting any of the radio buttons
would cause all other question bank radio input buttons to deselect.
This fix uses the question bank's id to differentiate it from others.


iturgeon:
fixes #1049

iturgeon:
I added a few optimizations, I'm pretty happy with it now.   I can't approve it since it's mine, but I think it's ready to go now.

PR #1051, opened 2020-01-02
Module Edit Locking

### LINKED ISSUE #1011: Lock editing to prevent multiple users overwriting each other
Now that we have the ability to have multiple users owning a module, we'll need to lock editing to prevent multiple users from overwriting each other's work.

In Obojobo Classic we just used a key in memcache to keep track of this - `user x is editing document y` (with a 5 minute auto-expire).  The key is refreshed every time the session ping occurs.

If another user attempts to edit a document, the ui blocks the action and tells them that Prof. X is editing it right now.

Make sure to deal with the chance that your own current editing rights can expire while you've got a tab backgrounded.

iturgeon:
Fixes #1011 

Adds ability to lock module to prevent multiple users from editing it at the same time.

## How Editor Locking works:

I'm actually pretty happy with this implementation - It should be very effective and unobtrusive.

`autoExpireMinutes` - If a user doesn't re-new it's lock within this time, other editors will be able to establish their own lock.  The editor will renew this lock automatically while it's open till idle time excedes idleMinutes.  The editor will attempt to delete it's lock when it's closed.

`warnMinutes` - The time before the editor warns the user about going idle.  A modal message pops up. As soon as the editor receives focus the modal automatically closes. The editor will continue to renew locks between warnMinutes and idleMinutes.

`idleMinutes` - Time before the editor stops renewing it's lock and actively removes it's own lock.  A modal message is displayed when this timer is hit.  If the ediotr receives focus after this time, it will attempt to re-establish a lock.  If posible, the modal will automatically close and the editor will continue where it left off.  If not - the editor will be locked with a message telling the author that someone else is editing.

Screen Shot 2020-06-15 at 3 49 14 PM


packages/app/obojobo-express/routes/api/locks.js
```diff
@@ -0,0 +1,53 @@
...
+router
+	.route('/:draftId')
+	.get([requireDraftId, requireCanViewEditor, checkValidationRules])
+	.get(async (req, res) => {
+		const existingLock = await EditLock.fetchByDraftId(req.params.draftId)
+		res.success(existingLock)
+	})
+
+// CREATE A LOCK
+// // mounted as /api/locks/:draftId
```
zachberry:
```suggestion
// mounted as /api/locks/:draftId
```

packages/app/obojobo-express/models/edit_lock.js
```diff
@@ -0,0 +1,87 @@
...
+			WHERE created_at < now() - interval '${editLockExpireMinutes} minutes'
+				`,
+			null,
+			r => r.rowCount // extract rowCount from query data
+		)
+	}
+
+	// clear a user's locks on a draft
+	// returns true if deleted, false if nothing to delete
+	static deleteByDraftIdandUser(userId, draftId) {
```
zachberry: It seems that this isn't being called, however I think we should as a `onbeforeunload` in the editor. We can use `navigator.sendBeacon` to post to an API end point to explicitly delete a lock when we know the user is stopping editing. We could potentially also have a `deleteByUser` which could be called on logout?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
```diff
@@ -123,6 +126,30 @@ class EditorApp extends React.Component {
...
+			}
+		})
+	}
+
+	startRenewEditLockInterval() {
+		setInterval(() => {
+			this.createEditLock(this.state.draftId).catch(() => {
+				this.displayLockedModal()
+			})
+		}, RENEW_LOCK_INTERVAL)
```
zachberry: Does this mean a user could leave a tab open and have a lock on a module indefinitely? Do we need to implement an inactive timeout to stop someone from doing this?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
```diff
@@ -123,6 +126,30 @@ class EditorApp extends React.Component {
...
+				title: 'Module is Being Edited.',
+				message: 'Someone else is currently editing this module, please try again later.'
+			}
+		})
+	}
+
+	startRenewEditLockInterval() {
+		setInterval(() => {
+			this.createEditLock(this.state.draftId).catch(() => {
+				this.displayLockedModal()
```
zachberry: This is a catch all for every error. If someone no longer has permissions to a module then they are shown this error, when in reality the error they are getting is `"Authoring permissions required to create a lock"`.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
```diff
@@ -30,6 +30,7 @@ const { OboModel } = Common.models
 
 const XML_MODE = 'xml'
 const VISUAL_MODE = 'visual'
+const RENEW_LOCK_INTERVAL = 60000 * 4.9 // 4.9 minutes in milliseconds
```
zachberry: Is it possible to pull this from the new config variable `editLockExpireMinutes`?

packages/app/obojobo-express/models/edit_lock.js
```diff
@@ -0,0 +1,87 @@
...
+		`,
+				{ draftId }
+			)
+			.then(data => {
+				if (data === null) return
+				return new EditLock(data)
+			})
+	}
+
+	// inserts a lock only when a differnt user's valid lock doesn't already exist
```
zachberry:
```suggestion
	// inserts a lock only when a different user's valid lock doesn't already exist
```

packages/app/obojobo-express/models/edit_lock.js
```diff
@@ -0,0 +1,87 @@
...
+	static create(userId, draftId) {
+		return db.one(
```
zachberry: Should this return a EditLock instance?

packages/app/obojobo-express/routes/api/locks.js
```diff
@@ -0,0 +1,53 @@
...
+const { checkValidationRules, requireDraftId, requireCanViewEditor } = oboRequire(
+	'express_validators'
+)
+
+// CHECK A LOCK
+// mounted as /api/locks/:draftId
+router
+	.route('/:draftId')
+	.get([requireDraftId, requireCanViewEditor, checkValidationRules])
+	.get(async (req, res) => {
```
zachberry: Should we call `EditLock.deleteExpiredLocks()` here? Otherwise it seems old locks never get cleared out unless someone tries to create one.

packages/app/obojobo-express/models/edit_lock.js
zachberry: Do we also need to call this one someone loses permissions to an object? If you start editing a module, then get your access to that module removed the lock remains and the original owner is stuck waiting for the lock to expire.

iturgeon: All good points, but I think having minimal impact right now, perhaps v2 of locking?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
iturgeon: That's true.  Since this is now the first thing to check your access rights in the editor it'll need to tell you if you just don't have permission.  I might be able to use the status to determine this

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
iturgeon: Yea.  even backgrounded tabs will keep firing.  We could udse a cheap solution for now - perhaps using the editor events?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
iturgeon: it'd be a lie to do so.  This gets compiled and thrown into js and the other is re-configurable at run time. We could change it so it's loaded from the server in the template launch.

zachberry:
IRL Discussion: We could notify users if someone else is editing a document to warn them, however we wouldn't stop them from editing. That way we solve the problem of someone locking a module for a very long period of time.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
```diff
@@ -62,7 +59,7 @@ class CodeEditor extends React.Component {
 
 	componentDidMount() {
 		// Setup unload to prompt user before closing
-		window.addEventListener('beforeunload', this.checkIfSaved)
+		// window.addEventListener('beforeunload', this.checkIfSaved)
```
qwertynerd97: Did we want to permanently remove the checkIfSaved ?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/page-editor.js
```diff
@@ -104,7 +103,7 @@ class PageEditor extends React.Component {
 
 	componentDidMount() {
 		// Setup unload to prompt user before closing
-		window.addEventListener('beforeunload', this.checkIfSaved)
+		// window.addEventListener('beforeunload', this.checkIfSaved)
```
qwertynerd97: Same as above, do we want to remove checkIfSaved?

packages/app/obojobo-express/server/models/edit_lock.js
```diff
@@ -0,0 +1,113 @@
...
+			WHERE created_at < now() - interval '${editLockExpireMinutes} minutes'
+				`,
+			null,
+			r => r.rowCount // extract rowCount from query data
+		)
+	}
+
+	// clear a user's locks on a draft
+	// returns true if deleted, false if nothing to delete
+	static deleteByDraftIdandUser(userId, draftId) {
```
zachberry:
```suggestion
	static deleteByDraftIdAndUser(userId, draftId) {
```

packages/app/obojobo-express/server/routes/api/locks.js
```diff
@@ -0,0 +1,69 @@
...
+	})
+
+// DELETE MY LOCK
+// would like to use delete method, but were relying on navigator.sendBeacon which is only post
+// mounted as /api/locks/:draftId/delete
+router
+	.route('/:draftId/delete')
+	.post([requireDraftId, requireCanViewEditor, checkValidationRules])
+	.post(async (req, res) => {
+		EditLock.deleteByDraftIdandUser(req.currentUser.id, req.params.draftId)
```
zachberry:
```suggestion
		EditLock.deleteByDraftIdAndUser(req.currentUser.id, req.params.draftId)
```

packages/app/obojobo-express/server/routes/api/locks.js
```diff
@@ -0,0 +1,69 @@
...
+			if (existingLock && existingLock.userId !== req.currentUser.id) {
+				return res.notAuthorized('Draft is already locked by someone else.')
+			}
+
+			const myLock = await EditLock.create(
+				req.currentUser.id,
+				req.params.draftId,
+				req.body.contentId
+			)
+
```
zachberry: What about this - instead of 

```javascript
existingLock = EditLock.fetchByDraftId(...)
if(existingLock and existingLock.userId is not me) throw 'Already locked'
myLock = EditLock.create(...)
```

We do

```javascript
myLock = EditLock.create()
if(myLock is empty or false or the userId is not my userId) throw 'Already locked'
```

Since EditLock.create has code to prevent anything from being inserted if a lock already exists (unless I'm mis-understanding the query)

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-app.js
```diff
@@ -123,6 +161,47 @@ class EditorApp extends React.Component {
...
+					)
+				}, this.editLocks.autoExpireMs)
+
+				this._isCreatingEditLock = false
+			})
+			.catch(() => {
+				this.displayLockedState()
+				this._isCreatingEditLock = false
+			})
+	}
```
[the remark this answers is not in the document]
iturgeon: I gave it a shot but had some issues w/ it passing tests, it was weirdly clearing when running the 2nd time.  I'm scared so I'm keeping it.  I also simplified a little - so I it might be a tad better

packages/app/obojobo-express/server/models/edit_lock.js
```diff
@@ -0,0 +1,118 @@
...
+			)
+
+			// verify the newest version is the same as the version requested
+			// if it is not - that means the draft has been updated unexpectedly
+			if (currentContentId.contentId !== contentId) {
+				throw Error('Current version of draft does not match requested lock.')
+			}
+
+			return await t
+				.one(
```
FrenjaminBanklin, line 94: Changing to `oneOrNone` resolves the error Zach described above, where attempting to edit a locked module throws an error instead of indicating that the module is locked:
```suggestion
				.oneOrNone(
```

Looks like the issue Elli mentioned is also still in play - would it be a significant imposition to clear the lock when the window unloads?

iturgeon:
The problem about unlocking on close - is that it's unreliable.  SendBeacon was supposed to solve this but: https://volument.com/blog/sendbeacon-is-broken

iturgeon:
We currently are using sendBeacon, but it isn't reliable either - so other editors may have to wait till the lock expires (the issue Ellie mentioned and Brandon verified)

iturgeon:
there's breadcrumbs to a solution in the link I posted above that point at using `visibilitychange` and `pagehide`

iturgeon:
@zachberry @FrenjaminBanklin I'm Excited - looks like I was able to get window close event much more reliable.  It needs some testing by you guys, but this should make the editor unlocking AND the viewer close event tracking more reliable!

packages/app/obojobo-express/server/models/edit_lock.js
iturgeon: Resolved, used Brandon's suggestion but needed a little more integration.

iturgeon:
@zachberry  I think you're right on the save solution.  It does look like the save method in editor-app was never wired up or wasn't completely unwired.  At either rate, locking needs to make sure the contentId is updated in the in-memory model.  That and there's some repetitive stuff (check for errors, display a modal) that we want to happen in both cases.

So I'm going to use the save method prop and have both editors use it to actually post the data to the api.

iturgeon:
I _think_ that covers everyone's issues.  🤞

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
```diff
@@ -109,17 +106,9 @@ class CodeEditor extends React.Component {
...
-
-				this.setState({ saved: true })
-			})
-			.catch(e => {
-				if (e instanceof Error) e = e.message
-				ModalUtil.show(<SimpleDialog ok title={`Error: ${e}`} />)
-			})
+		return this.props.saveDraft(draftId, code, mode).then(isSaved => {
+			this.setState({ saved: isSaved })
+		})
```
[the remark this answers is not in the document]
iturgeon: This is handled on #1158  https://github.com/iturgeon/Obojobo/blob/issue/228-score-importing/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js#L112-L124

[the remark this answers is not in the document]
iturgeon: I'm going to mark this resolved as it's handled elsewhere

zachberry:
Closing in favor of #1461

PR #1055, opened 2020-01-06
Fixes issues when clicking on Break nodes or text placeholders causing annoying behavior

zachberry:
* Fixed an issue where placeholders ("Type your text here") were blocking users from being able to click on the document in FF
* Turned Break into a non-void node with a 'ghost' text element, preventing an issue where clicking or focusing in on the break node would cause the scroll to jump to a random area of the page
* Fixed the same issue when clicking on the button to toggle the size of the break
* Updated the styling of the break in the editor when full size to match the way it displays in the editor
* Added white borders around the insert menu dotted line and more info buttons so that elements that intersect it have some visual padding


zachberry:
Fixes #1054

zachberry:
Closing in favor of #1063

PR #1057, opened 2020-01-06
Fixes a few issues with copy & paste in the visual editor

zachberry:
* Update registry.registerEditorModel to use defaults if any given EditorNode methods are not set (previously these would remain as `undefined`)
* Added a Heading onPaste listener to insert plain text fixing an issue where the whole page would crash
* Corrected variable name from an earlier code-copy-and-paste leftover in Text editor-registration (changed isList to isText)


packages/obonode/obojobo-chunks-heading/editor-registration.js
```diff
@@ -18,6 +23,21 @@ const Heading = {
...
 	plugins: {
+		onPaste(event, editor, next) {
+			const isHeading = isType(editor)
+			const transfer = getEventTransfer(event)
+
+			// If we are pasting standard text (not-html) then we simply
+			// put the text inside the existing heading node. If we don't
+			// do this the page will crash when pasting text into a blank
+			// heading node.
+			if (isHeading && transfer.type === 'text') {
```
qwertynerd97:
```suggestion
			if (isHeading && transfer.type !== 'fragmentt') {
```

qwertynerd97: To prevent a crash from html

iturgeon: Can confirm this prevents pasted html into the heading from crashing

iturgeon:
```suggestion
			if (isHeading && transfer.type !== 'fragment') {
```
fixes typo in @qwertynerd97 's suggestion above 'fragmentt'

PR #1061, opened 2020-01-07
Ensures that triggers in the Visual Editor now work for every node type, and converting from/to XML maintains trigger information

### LINKED ISSUE #1056: Switching from visual editor to XML removes nav lock functionality
As reported:

```
Bug report: The xml editor breaks navlocks. It looks like the xml editor will remove the lock navigation during attempts because it doesn't write it in the xml

So if I open the xml editor, do nothing else but hit save file, the locking mechanism is broken.
new messages
json editor doesn't break it, I'm assuming that it is because the visual editor directly writes json, and it is the translator that is malfunctioning.
```

zachberry:
This PR requires #746 !

Fixes #1056 

* Fixed a bug where the JSON->XML parser didn't output the ID of Break OboNodes
* JSON->XML parser now outputs triggers for all OboNodes rather than just the ActionButton
* Updated editor converter.js files to include triggers and to remove undefined attributes
* Added missing ActionButton node parser jest in the JSON->XML parser package
* Added utility function, common/util/without-undefined.js, which removes undefined values from an object
* Updated registry to use the without-undefined method


zachberry:
A note for myself, this seems to be causing the rubric mods dialog not to open

PR #1063, opened 2020-01-08
Fixes scrolling focus issues for Break, MathEquation and YouTube in the Visual Editor

### LINKED ISSUE #1062: Fix MathEquation and YouTube from randomly scrolling on page when clicking on them in the Visual Editor
Similar issue to #1052. That issue solves the problem for Break nodes which was reported by an end user but I observed that the same problem exists for MathEquation and YouTube. Will double check that their aren't any other nodes having the same issue as well.

zachberry:
This PR includes #1055 

Fixes #1062 and #1054 

* Adds EditableHiddenText component as a standard way to render a dummy contentEditable slate text node
* Adds includeTextCancellingPlugins as a standard way to define plugins for slate nodes which have a dummy text node
* Updates MathEquation and YouTube to work like Break with a dummy text node
* Updates Break to use new standard dummy text features
* MathEquation changed to update slate state only when editing dialog is closed
* Focuses on Latex input when opening the MathEquation edit dialog


packages/app/obojobo-document-engine/src/scripts/oboeditor/util/include-text-cancelling-plugins.js
```diff
@@ -0,0 +1,37 @@
...
+		return next()
+	}
+
+	// @HACK: There isn't a way to prevent pasting in this version of
+	// slate, so if the user is focused on our blank node and pastes
+	// we need to cancel it. The hack du jour is undoing the paste action.
+	pluginsConfig.onPaste = (event, editor, next) => {
+		if (isType(editor, nodeType)) {
+			next()
+			editor.undo()
```
qwertynerd97: These two cancel each other out and can be removed without negatively impacting the editor

PR #1073, opened 2020-01-10
Prevent QuestionBank settings component from creating infinite loop

### LINKED ISSUE #1071: QuestionBank Settings componentDidUpdate causes infinite loop
Noticed this after merging in #1050 - The componentDidUpdate method causes updateNodeFromState which in turn calls componentDidUpdate and so on. The debounce caused the page not to halt so it wasn't noticed.

Solution is to run updateNodeFromState only if component state differs from slate node state.

zachberry:
Fixes #1071


zachberry:
Re-reviewed, looks good to me!

Issue #1070, opened 2020-01-10
Editors not always saving as expected


zachberry:
Found 3 cases where the editors are not saving as expected.

1. In the visual editor if you rename a module the 'Saved!' text shows up but the module isn't actually saved.
2. If you switch to the XML editor the module is automatically saved, but if you switch to the Visual Editor it's not, and if you haven't saved any XML work is lost. I assume the same problem exists for the JSON editor but I haven't tested it.
3. In the XML editor if you have invalid XML and attempt to save you should see an error dialog pop up. However if you use the keyboard shortcut (cmd/ctrl+s) to save you'll see 'Saved!' with no indication that something went wrong.

vutoan1245:
For point number 2: `onSave()` was called but it is an asynchronous function. As a result, the editor switches before it is done saving. (File reference: `obojobo-document-engine/src/scripts/oboeditor/components/toolbars/view-menu.js`)

iturgeon:
I just noticed when I fetched all the repos that you committed some code for this.  I went pretty deep into trying to understand & simplify how editor-app handles it's 3 different states - ended up in some wide-sweeping changes on my branch https://github.com/ucfopen/Obojobo/pull/1079

PR #1082, opened 2020-01-13
Fix save problems on visual/xml/json editor

vutoan1245:
Overview:
1. `File` -> `Rename` -> `OK`: This will rename and save the draft
2. Fixed problem draft does not save when switching between XML, JSON, and Visual editor.
3. Fixed problem on when `cmd/ctrl+s` does not prompt error to the user.

Resolve: #1070

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
@@ -12,10 +12,12 @@ import 'codemirror/addon/fold/foldgutter.css'
 import 'codemirror/addon/fold/xml-fold.js'
 import { Controlled as CodeMirror } from 'react-codemirror2'
 
-import APIUtil from 'obojobo-document-engine/src/scripts/viewer/util/api-util'
-import EditorUtil from 'obojobo-document-engine/src/scripts/oboeditor/util/editor-util'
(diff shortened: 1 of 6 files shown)


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
```diff
@@ -16,6 +16,8 @@ import APIUtil from 'obojobo-document-engine/src/scripts/viewer/util/api-util'
 import EditorUtil from 'obojobo-document-engine/src/scripts/oboeditor/util/editor-util'
 import FileToolbar from './toolbars/file-toolbar'
 import hotKeyPlugin from '../plugins/hot-key-plugin'
+import ModalUtil from 'obojobo-document-engine/src/scripts/common/util/modal-util'
+import SimpleDialog from 'obojobo-document-engine/src/scripts/common/components/modal/simple-dialog'
```
iturgeon: These should be relative since we're in the obojobo-document-engine package: `../../common/components/modal/simple-dialog`

iturgeon: for that matter, line 16 should be updated too

zachberry:
I found #1083 but it's also a problem on dev/8 and I don't think it's high priority so spun it into a different issue.

PR #1084, opened 2020-01-13
Issue/1026 module sort options

### LINKED ISSUE #1026: Add sorting options to the Repository modules page
Need to be able to sort this list alphanumerically or by date. Preferably we could keep the preference stored.

ctcuff:
Adds a sorting option to the repository modules page. Resolves #1026.

### DIFF
--- a/packages/app/obojobo-repository/server/routes/dashboard.js
+++ b/packages/app/obojobo-repository/server/routes/dashboard.js
@@ -14,10 +14,20 @@ router
 	.route('/dashboard')
 	.get([requireCurrentUser, requireCanPreviewDrafts])
 	.get((req, res) => {
+		const cookies = req.headers.cookie.split(';')
+		let selectedSort = 'Alphabetical'
+
+		cookies.forEach(cookie => {
+			if (cookie.includes('selectedSort')) {
+				selectedSort = cookie.split('=')[1]
+			}
+		})
+
 		return DraftSummary.fetchByUserId(req.currentUser.id).then(myModules => {
 			const props = {
 				title: 'Dashboard',
 				myModules,
+				selectedSort,
 				currentUser: req.currentUser,
 				appCSSUrl: webpackAssetPath('dashboard.css'),
 				appJsUrl: webpackAssetPath('dashboard.js')
--- a/packages/app/obojobo-repository/server/routes/dashboard.js
+++ b/packages/app/obojobo-repository/server/routes/dashboard.js
@@ -14,20 +14,19 @@ router
 	.route('/dashboard')
 	.get([requireCurrentUser, requireCanPreviewDrafts])
 	.get((req, res) => {
+		let sortOrder = 'alphabetical'
 		const cookies = req.headers.cookie.split(';')
-		let selectedSort = 'Alphabetical'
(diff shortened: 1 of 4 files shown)


packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -65,62 +65,113 @@ const renderModalDialog = props => {
...
-					<div className="repository--item-list--collection">
-						<div className="repository--item-list--collection--item-wrapper">
-							<div className="repository--item-list--row">
-								<div className="repository--item-list--collection--item--multi-wrapper">
-									{renderModules(props.filteredModules ? props.filteredModules : props.myModules)}
+		const selectedSort = props.selectedSort
+
+		this.state = {
+			selectedSort,
+			sortOptions: ['Newest', 'Alphabetical', 'Last updated']
```
iturgeon: Don't keep options like these in state - their constants for the 'class'  either create a const in the scope somewhere or make them a property of the class.

packages/app/obojobo-repository/server/routes/dashboard.js
```diff
@@ -14,10 +14,20 @@ router
...
 	.get([requireCurrentUser, requireCanPreviewDrafts])
 	.get((req, res) => {
+		const cookies = req.headers.cookie.split(';')
+		let selectedSort = 'Alphabetical'
+
+		cookies.forEach(cookie => {
+			if (cookie.includes('selectedSort')) {
+				selectedSort = cookie.split('=')[1]
+			}
+		})
```
iturgeon: Suggest using `array.find` since we're trying to pluck out a single value and don't need to continue looping after locating it.

```javascript
const cookieSort = cookies.find(cookie => {
  if(cookie.includes('selectedSort'){
    return cookie.split('=')[1]
  }
})
```

But before that, look at`cookies.get` https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/get  I think we can use it here instead.

packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -65,62 +65,113 @@ const renderModalDialog = props => {
...
+		switch (this.state.selectedSort) {
+			case 'Alphabetical':
+				modules.sort((a, b) => a.title.localeCompare(b.title))
+				break
+			case 'Newest':
+				modules.sort((a, b) => new Date(b.createdAt) - new Date(a.createdAt))
+				break
+			case 'Last updated':
+				modules.sort((a, b) => new Date(b.updatedAt) - new Date(a.updatedAt))
+				break
```
iturgeon: Not a required change, but I'd try pulling the call to `modules.sort` out of the switch, instead store the sorting function in a variable like `sortFn` then call `modules.sort(sortFn)` after the switch.

Possible reasons why: it's nice to limit what happens in the logical forks in the code to just what's different. And it makes it clear that all paths lead to sorting.

That said, like any polishing step - try it out and if it's not any better don't do it.

iturgeon:
Another thing to check on, the upcoming changes to SameSite defaults for cookies may affect obojobo when it's in an iframe

PR #1105, opened 2020-01-22
Update variable substitution code to not break Latex

### LINKED ISSUE #1104: {{ characters stripped from LaTeX equations causing them not to render
From a user submitted report:

1. Insert this text in the editor: `\displaystyle{v_f = \sqrt{{v_{f,x}}^2 + {v_{f,y}}^2}}`
2. Select it and style it as a Math Equation
3. Preview the module - note that the page doesn't render

### LINKED ISSUE #1139: Bad in-line latex equation crashes Viewer
Creating a bad inline latex equation such as `\frac{` will completely crash the Viewer - Instead bad equations should render with some sort of error state so authors can see that their equation is not working.

zachberry:
Fixes #1104
Fixes #1139

This getText method now won't replace the `{{` or `}}` characters if the variable inside of them is not recognized.

Before this change: `A {{unrecognized}} variable` -> `A unrecognized variable`
After this change: `A {{unrecognized}} variable` -> `A {{unrecognized}} variable`

I think this is expected behavior and authors will more easily see if they made a typo in a variable. This also solves the reported issue of breaking latex which includes double brackets.


ctcuff:
Not sure if it's related to this issue specifically, but entering unbalanced braces causes the preview to load forever. For example `\int_{0}^{2\pi}{{(sin^2x + cos^2x)}`

iturgeon:
Looks good to me.  Based on cam's feedback though, I thought I'd throw in a few changes to render invalid katex errors in the viewer.

In the Editor:
Screen Shot 2020-02-25 at 10 40 06 PM

In the Viewer:
Screen Shot 2020-02-25 at 10 39 58 PM

zachberry:
Looked at the updates and I'm happy with them, it all LGTM

Issue #1102, opened 2020-01-22
Pasting plain text into figure caption breaks visual editor


zachberry:
I get the error `Uncaught TypeError: Cannot read property 'size' of undefined`, seems to be attempting to replace the node in the same way that pasting into headings was causing issues.

PR #1107, opened 2020-01-23
Pasting text in figure caption breaks editor

vutoan1245:
Overview:
- Add onPaste listener to insert plain text into Figure
- Add safety feature to prevent Figure from breaking

Similar PR: #1057
Resolve: #1102


iturgeon:
I had to cherry pick the changes onto master since we want to target this onto master (rebase was a little too gnarly)

Issue #1059, opened 2020-01-06
File -> New loads 404 page in XML and JSON editors


zachberry:
Want to double check that other items in the menus don't cause problems in the XML and JSON editors.

PR #1116, opened 2020-01-27
Creating a new module in xml/json editor causes 404

vutoan1245:
When creating a new module in XML editor, the new module is launched into `/editor/xml/8a51d...`
Change `xml` to `visual` fixes the problem.

Resolve: #1059 
Fix the same problem for `Make a copy`


vutoan1245:
@zachberry As mention on Slack, I will fix that issue in a separate PR.

Issue #1129, opened 2020-01-30
'Make a copy' working unexpectedly


vutoan1245:
1. When `Make a copy` in Visual Editor, both titles of the current document and the new document are changed.
2. a problem found by @zachberry #1116
> Noticed one thing - if you make a copy while in JSON or XML view it seems to give you a new blank object rather than a copy. For example, go to XML mode, change the title to 'blah', save, then make a copy - the new copy will be called the default instead of ("blah - Copy")
>

PR #1013, opened 2019-12-17
Copy a public module

vutoan1245:
Add feature to copy a module to the personal library.

Overview :
- Add a new database table to store metadata for a document: `drafts_metadata`
- Add a new API endpoint to copy the document: `/api/drafts/:draftId/copy/`

Resolve issue #1005

PR #1132, opened 2020-01-30
`Make a copy` feature is not working as expected

vutoan1245:
Resolve: #1129 

I am opening to suggestions

Problem 1:
- Reason: The current document is renamed the then it is copied to the new document.
- Solution: Rename the current document back after copying. This feels a bit hacky but simple. Users can see the title got change for a brief moment.

Problem 2:
- Reason: Code Editor and Visual Editor are using different `onSave` function. The `onSave` of Visual editor takes in `draftId` and saves the document to that `draftId`. The `onSave` of the Code Editor saves to the current draftId.
- Solution: Make `onSave` of the Code Editor acts similarly to the `onSave` of the Visual Editor.

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/code-editor.js
@@ -113,14 +113,14 @@ class CodeEditor extends React.Component {
 		})
 	}
 
-	saveCode() {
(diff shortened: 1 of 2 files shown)


vutoan1245:
Another solution is to use the copy feature from https://github.com/ucfopen/Obojobo/pull/1013. One advantage is that the metadata is saved to the database.

PR #1134, opened 2020-01-30
[update] cookie to support sameSite=none

### LINKED ISSUE #1119: Chrome 80 SameSite Cookie changes
We'll need to update the cookie settings to allow SameSite=None for people who don't host Obojobo and Canvas on the same root domain.

Reading: 
https://web.dev/samesite-cookies-explained/
https://community.canvaslms.com/docs/DOC-18313-samesite-cookies-and-canvas

iturgeon:
fixes #1119

### DIFF
--- a/packages/app/obojobo-express/middleware.default.js
+++ b/packages/app/obojobo-express/middleware.default.js
@@ -43,10 +43,12 @@ module.exports = app => {
 			rolling: true,
 			cookie: {
 				path: '/',
-				sameSite: false, // Seems to be blocking access from webcourses2c.instructure.com to obojobo (though not webcourses.ucf.edu?)
+				maxAge: 10 * 24 * 60 * 60 * 1000, // 30 days
+				// use config.general.secureCookie to determine the next settings
+				// sameSite 'none' is needed to support LTI launches from different domains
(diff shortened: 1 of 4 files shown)


packages/app/obojobo-express/middleware.default.js
```diff
@@ -43,10 +43,12 @@ module.exports = app => {
 			rolling: true,
 			cookie: {
 				path: '/',
-				sameSite: false, // Seems to be blocking access from webcourses2c.instructure.com to obojobo (though not webcourses.ucf.edu?)
+				maxAge: 10 * 24 * 60 * 60 * 1000, // 30 days
+				// use config.general.secureCookie to determine the next settings
+				// sameSite 'none' is needed to support LTI launches from different domains
+				sameSite: config.general.secureCookie ? 'none' : false,
```
zachberry: Do we maybe want a different config value for this instead of going with secureCookie?

iturgeon: Just as a note for the future - sameSite=none requires the cookie to be secure, so they're somewhat linked together.  The only situation I could imagine wanting a separate flag for this is when you are hosting your LMS and obonext on the same domain and would like to lock down that cookie a little more.

PR #1149, opened 2020-02-10
Bad latex no longer crashes the viewer

### LINKED ISSUE #1139: Bad in-line latex equation crashes Viewer

ctcuff:
Fixes #1139 

Looking for some opinions on on this one, wondering if there's a better way to display an error. Right now it renders like this:
katex-error-no-color

I think it might be more noticeable with color:
latex-error-with-color


iturgeon:
Good work, but it looks like fixes over on #1105 have this fixed.

PR #1153, opened 2020-02-11
Convert scoreActions to/from attrs to newer for attr when editing in the visual editor

### LINKED ISSUE #1150: Visual Editor strips deprecated 'from' and 'to' attrs for scoreActions
To replicate:
1. Create an XML or JSON document with a score action with `from` and `to` properties, save.
2. Edit with Visual Editor and note that the Score Range is now undefined. These properties get filtered out.

If saved this document will break the viewer.

We need to either convert these values in the editor converter or support them and not filter them out.

zachberry:
Resolves #1150


ctcuff:
Looks like it's working to me, but if you edit in JSON or XML, you can save so that `from` > `to`. Not sure if that would actually cause any problems, but I think the visual editor should try and make sure the values are valid (maybe swap if `from` is larger then `to`)

PR #1160, opened 2020-02-19
XML editor now preserves unexpected tags and their attributes instead of breaking.

### LINKED ISSUE #1151: Invalid styles in XML break editor
To replicate:

1. Write this XML: `<p>Test <u>underline</u></p>`. (We don't support `u`)
2. Save, refresh page
3. Go back to XML edit mode
Result: The page crashes when attempting to parse the underline style

Not sure yet how to solve this - Best thing is to warn the user that they put in an invalid style. Or, we could allow unsupported styles and maintain the <u> even though it won't show up in the viewer. Or, probably the worse option, we could filter the u style out.

FrenjaminBanklin:
Fixes #1151.

Adjusts style range normalization and rendering processes to enable end users to use previously unsupported (or arbitrary) XML tags in the XML editor. Unsupported tags will be preserved and displayed in the XML (and JSON) editor(s) between edits.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/common/text/chunk-style-list.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/common/text/chunk-style-list.test.js
@@ -302,6 +302,46 @@ describe('ChunkStyleList', function() {
 		])
 	})
 
+	test('normalizes unexpected styles', () => {
+		styleList = new ChunkStyleList()
+
+		styleList.add(new StyleRange(0, 5, 'b', {}))
+		styleList.add(new StyleRange(0, 5, 'extra', {}))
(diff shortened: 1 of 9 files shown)


FrenjaminBanklin:
Re: conversations;

It sounds like what's in this branch is minimally viable and good enough to push through because it meets the need which is 'keep the XML Editor from crashing or throwing away unsupported tags.'

Opening up a document in the viewer with unsupported tags will render those tags, which throws console errors but generally still works (will open an issue for that). Potentially that's worth addressing. The more pressing concern is that opening the document in the Visual Editor will strip any unsupported tags (will open an issue for that).

Issue #1126, opened 2020-01-30
Import / upload a module via document


zachberry:
You can download documents as XML or JSON files, but we should allow you to import / upload these instead of having to manually copy and paste

PR #1168, opened 2020-03-03
Add feature to import local module

vutoan1245:
There are two ways to import a local Obojobo document:
1. In the Dashboard: `New Module` -> `Import From File`
2. In the Editor: `File` -> `Import from file...`

Resolve: #1126


vutoan1245:
@FrenjaminBanklin 

The reason there are two versions of the document is that I was lazy and reused old APIs
Step 1. call `APIUtil.createNewDraft` to create a blank module
Step 2. call `APIUtil.postDraft` to replace the blank module with the content of the select file

Not sure if it is a good idea to do it that way. I feel like I should create a new API for this feature.
I am opening to suggestions.

zachberry:
@vutoan1245 I think perhaps we could update /api/drafts/new to accept content, which would then create the new draft with the given content. If none was provided then it would default to the draftTemplate and draftTemplateXML values.

keeeeeegan:
> The UI doesn't make it very clear that importing from file makes a new module instead of overriding the content of the current system. Maybe move the import button to below new and label it "Import New from File"?
> 
> Also, because the download for JSON currently includes staus codes, users who download a file as JSON and then re-upload it receive a confusing error message "Cannot read property 'start' of undefined"

I'm seeing this behavior as well

iturgeon:
I was having a hard time fixing the conflicts so I just made a new commit w/ the changes Toan did (- test updates) here https://github.com/iturgeon/Obojobo/commit/cbebea749786660120cf5fe1433f06615e348ab9

iturgeon:
welp, that's terrible

packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -106,6 +107,33 @@ const Dashboard = props => {
...
+	const onChangeFile = event => {
+		const file = event.target.files[0]
+		if (!file) return
+
+		const reader = new FileReader()
+		reader.readAsText(file, 'UTF-8')
+		reader.onload = function(e) {
+			const contents = e.target.result
+
+			APIUtil.createNewDraft(
```
iturgeon, line 119: This should happen via a redux in an action.

packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -106,6 +107,33 @@ const Dashboard = props => {
...
+				contents,
+				file.type === 'application/json' ? 'application/json' : 'application/xml'
+			).then(() => {
+				location.reload()
+			})
+		}
+	}
+
+	const importFile = () => {
+		const fileSelector = document.createElement('input')
```
iturgeon, line 129: This should probably also be in an action

packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -12,6 +12,7 @@ const Button = require('./button')
 const MultiButton = require('./multi-button')
 const Search = require('./search')
 const ReactModal = require('react-modal')
+const APIUtil = require('obojobo-document-engine/src/scripts/viewer/util/api-util')
```
iturgeon, line 15: I'd prefer not to use the viewer's api-util here, it has a lot more then we need in it

iturgeon:
Ok, this is ready again

vutoan1245:
I really like the changes you added @iturgeon. It is working great.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/page-editor.js
```diff
@@ -46,10 +46,12 @@ class PageEditor extends React.Component {
 			showPlaceholders: true
 		}
 
+		this.imported = false
```
iturgeon, line 49: Looking at the diff, this may not be needed any more?

vutoan1245: True

keeeeeegan:
LGTM, tested locally

PR #1173, opened 2020-03-03
Xml encode special character

### LINKED ISSUE #1141: Converting & to &amp; in XML (and other special characters)
To replicate:
1. In the visual editor type an ampersand
2. Switch to XML editing
3. Attempt to save the document

Result: Saving will fail since the & should be written as &amp;

We need to somehow convert this character and other special characters that need to be handled in XML when switching from XML, and restore the non-special version when switching away from XML

vutoan1245:
Resolve #1141 
Note: remember to run `yarn install`


qwertynerd97:
Similarly, the content inside latex seems to be double escaped.  If there is a < or > symbol in the latex, it gets escaped, and then the & in the escape character gets escaped again.
Screen Shot 2020-03-04 at 1 50 40 PM

Screen Shot 2020-03-04 at 1 49 15 PM

PR #1190, opened 2020-03-13
Updates share dialog to include email address

### LINKED ISSUE #1189: Share list doesn't have a way to distinguish between users with the same name
We should add an email address or some other identifier to help distinguish between users with the same name.

zachberry:
Fixes #1189


zachberry:
This is on hold at the moment as it may pose a FERPA issue. Perhaps these could be changed to be Obojobo user ids, however we don't really expose that anywhere.

zachberry:
We're thinking of swapping out the email for the user ID (you can already search the dialog by user ID)

zachberry:
Email has been swapped out for the user ID

Issue #1171, opened 2020-03-03
Able to move nodes up and down the page or between pages


Zhongzhou:
Being able to move either a figure, an equation, a practice problem or even just a piece of text up or down on the same page, or onto different pages. If on the same page it is the same as re-ordering the nodes. Could be accomplished via a "move to" dialog box, or a pair of "move up"/"move down" buttons.

PR #1192, opened 2020-03-16
Add Move Up / Move Down buttons 

vutoan1245:
Screen Shot 2020-03-16 at 10 10 55 AM

Resolve issue #1171


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.js
```diff
@@ -241,14 +241,8 @@ class MoreInfoBox extends React.Component {
...
-								{this.props.isFirst ? null : (
-									<Button onClick={() => this.props.moveNode(this.props.index - 1)}>Move Up</Button>
-								)}
-								{this.props.isLast ? null : (
-									<Button onClick={() => this.props.moveNode(this.props.index + 1)}>
-										Move Down
-									</Button>
-								)}
+								<Button onClick={() => this.props.moveNode(this.props.index - 1)}>Move Up</Button>
+								<Button onClick={() => this.props.moveNode(this.props.index + 1)}>Move Down</Button>
```
qwertynerd97: Use the index to hide buttons that would not have an effect

vutoan1245:
New changes:

- Move Up and Move Down buttons are disabled when applicable
- You can no longer move an Assessment
- A node cannot move pass multiple-choice options in Question

vutoan1245:
Resolved conflict. Not sure why Prettier formatted many unrelated files.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/editor-component.js
```diff
@@ -106,10 +119,9 @@ class Node extends React.Component {
 				{selected ? (
 					<MoreInfoBox
 						className="content-node"
+						index={index}
+						type={this.props.node.type}
```
iturgeon, line 123: it looks like `this.props.node.type` should be `this.props.element.type`

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/editor-component.js
```diff
@@ -78,7 +78,20 @@ class Node extends React.Component {
...
 	}
 
+	moveNode(index) {
+		const editor = this.props.editor
+
+		editor.moveNodeByKey(this.props.node.key, this.props.parent.key, index)
+	}
+
 	render() {
+		let size = this.props.parent.nodes.size
```
iturgeon, line 88: it looks like `this.props.parent` doesn't exist anymore

iturgeon:
Fixed up an ready again for testing.  Discovered a bug upstream in dev/10 #1266

zachberry:
Also, want to check that this will be compliant with #1212

vutoan1245:
It should be ready now. Not sure why it closes the second time but I think we can put into another issue.

PR #1195, opened 2020-03-18
Better define height of page to fix next/prev buttons from being cut off

### LINKED ISSUE #1194: Scroll position / height of pages off
Not sure when this ended up happening but it seems that the default position of the page is not working correctly.

Notice the Next button is cut off:

image

zachberry:
Fixes #1194


zachberry:
Note to self: Need to test case where the title is very very long, which would exceed the min-height of the button

zachberry:
Updated to force the buttons to be 6.5rem always instead of the min-height. Otherwise pages with not enough content to fill the available height was not positioning the buttons correctly.

PR #1159, opened 2020-02-19
Upgrade Slate to 0.5x

qwertynerd97:
Upgrades to the latest version of Slate, and adds in some quality of life fixes
This also includes the changes from #1008

PR #1207, opened 2020-03-23
Inline Latex Previews

### LINKED ISSUE #1206: Inline LaTeX previews
Played around a little with this idea - should greatly diminish authors writing bad inline latex:

image

zachberry:
Requires #1159 
Fixes #1206


keeeeeegan:
I noticed something. If your selection of text includes other formatting (such as sup text) and then click the function button, a blank page loads and you can't go back. I had this happen also for headings.



I also tried this with 9 and didn't see this issue.

iturgeon:
Something strange is going on w/ this PR - too many file changes (they include commits already in dev10)

qwertynerd97:
The weird repeated file changes have been fixed by merging emerald into the PR!

keeeeeegan:
Checked again, looks good!

zachberry:
Made a few changes:

* The preview wasn't showing up when (for example) put in Question explanations. Increased the z-index. Also moved where HoveringPreview exists in the JSX such that it will be under the toolbar at the top
* Previews weren't moving when scrolling the page - moving where the component is in the JSX also fixed this. Had to tweak the way it positioned itself, requiring a reference to the editor-page DOM element to get working
* When resizing the page the preview wasn't moving. Added a ResizeObserver to listen for this. ResizeObserver doesn't work in Safari, but the alternative is not worth it imho, and this is somewhat obscure and won't break anything if the preview doesn't move when resizing.

packages/app/obojobo-document-engine/__tests__/oboeditor/components/page-editor.test.js
```diff
@@ -25,8 +25,9 @@ jest.mock('src/scripts/oboeditor/stores/editor-store', () => ({
 jest.mock('src/scripts/oboeditor/components/navigation/editor-nav')
 jest.mock('src/scripts/oboeditor/components/toolbars/file-toolbar')
 jest.mock('src/scripts/oboeditor/components/toolbars/paragraph-styles')
-// jest.mock('src/scripts/oboeditor/components/toolbars/content-toolbar')
-//jest.mock('obojobo-document-engine/src/scripts/common/registry')
+jest.mock('src/scripts/oboeditor/components/hovering-preview', () => props => (
```
iturgeon, line 28: I added a mock for hovering-preview here to avoid rendering it in this test

packages/app/obojobo-document-engine/__tests__/oboeditor/components/hovering-preview.test.js
```diff
@@ -0,0 +1,176 @@
+import renderer, { act } from 'react-test-renderer'
+import React from 'react'
+import { Editor } from 'slate'
+import { ReactEditor, useSlate } from 'slate-react'
+import HoveringPreview from 'src/scripts/oboeditor/components/hovering-preview'
+
+jest.mock('slate-react')
+jest.mock('slate')
+
+describe('HoveringPreview', () => {
```
iturgeon, line 10: Added a test just for HoveringPreview

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/hovering-preview.js
```diff
@@ -0,0 +1,70 @@
+import React, { useRef, useLayoutEffect, useMemo } from 'react'
+import { ReactEditor, useSlate } from 'slate-react'
+import { Editor } from 'slate'
+import katex from 'katex'
+
+import './hovering-preview.scss'
+const LATEX_MARK = '_latex'
+
+const HoveringPreview = ({ pageEditorContainerRef }) => {
```
iturgeon, line 9: Named props

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/hovering-preview.js
```diff
@@ -0,0 +1,70 @@
...
+	const ref = useRef()
+	const editor = useSlate()
+	const [leaf] = editor.selection
+		? Editor.leaf(editor, editor.selection, { edge: 'start' })
+		: [{ text: '' }, []]
+
+	// SHOW, HIDE, AND UPDATE LOCATION ON SCREEN
+	// only runs when text changes OR window width changes
+	useLayoutEffect(() => {
```
iturgeon, line 18: Switched to `useLayoutEffect` since this changes the dom element's style

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/hovering-preview.js
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		// Calculate styles to set display location
+		const pageEditorRect = pageEditorContainerRef.current.getBoundingClientRect()
+		const rect = parent.getBoundingClientRect()
+
+		el.style.opacity = 1
+		el.style.top = `${rect.top - pageEditorRect.top - el.offsetHeight - 6}px`
+		el.style.left = `${rect.left - pageEditorRect.left - el.offsetWidth / 2 + rect.width / 2}px`
+	}, [window.innerWidth, leaf.text, leaf])
```
iturgeon, line 48: Added conditional watchers to useLayoutEffect - this should now only render when one of these changes

* window.innerWidth - no need to re-render on height changes as they have no effect on our calcs
* leaf.text - any time the leaf node text changes - this can be more often then desired because whatever is selected will feed it's own text in, latex or not
* leaf - if the leaf object itself is different (needed to recalc the location for 2 different equations w/ the same text)

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/hovering-preview.js
```diff
@@ -0,0 +1,70 @@
...
+
+	// RENDER KATEX HTML
+	// run only when text changes
+	const katexHTML = useMemo(() => {
+		if (!leaf[LATEX_MARK]) return ''
+		return katex.renderToString(leaf.text, { throwOnError: false })
+	}, [leaf.text])
```
iturgeon, line 55: katex rendering is fairly costly - so I added useMemo here only calculate when leaf text changes.

Another limitation - it'll run as the selection changes, even if its not a katex leaf.  So I added an additional quick exit inside the function if the mark isn't a latex mark

zachberry:
@iturgeon Updates change the behavior when you use the arrow keys to move into a inline latex equation. Before the preview would be positioned as expected:



With the updates clicking works but using the keyboard to move the cursor inside the equation results in the preview window showing up at the wrong location:

zachberry:
In hovering-preview switching `useLayoutEffect` for `useEffect` does fix the positioning problem, if that doesn't bog down our rendering too much

iturgeon:
Oh interesting, I obviously wasn't using arrow keys.  I'll have to think about the differences between useEffect and useLayoutEffect more

iturgeon:
Conflicts solved and switched back to useEffect after some reading.

iturgeon:
Hovering preview is above the more details menus.

Screen Shot 2020-04-16 at 10 23 30 AM

zachberry:
I can get the More Info Box to show up over the inline preview in basic nodes like texts and lists, but inside a Question (for example) the stacking contexts are different and the preview still shows up over the box.

I can hack around this by changing the focus into the MoreInfoBox:



It's also nice since keyboard focus is now inside the box. However, does this pose an issue with accessibility since the slate selection is lost? I was thinking we could capture the slate selection and restore it on close, but I'm not sure how (and if that is possible). Summoning @qwertynerd97

qwertynerd97:
Yeah, I actually just finished setting the more info box to focus on open and restore selection on close on my keyboard accessibility branch.  I've made a draft PR so you can take a look over at #1268

zachberry:
In this case, I'm thinking we could let the layer issue ride for now and fix when #1268 gets merged in

Issue #1142, opened 2020-01-31
Restore old versions


zachberry:
This would allow authors to restore an older version of a draft. This would be very helpful if something went wrong and a change to a document caused it to crash.

This would live in the repository on the 'Module Options' dialog and add a 'Restore' button. That would launch another modal which would show a list of older versions (newest to oldest) on the left side and a preview window on the right. Clicking an item in the list would let them preview that version. A button would also let them restore that version.

Restoring a module would mark the modules above it as deleted and flag them as deleted in the database, then create a new version which would sit on top of the list. The deleted items would simply be flagged as deleted, and could be restored and previewed as well.

zachberry:
The list also should say who saved the document. That would be helpful in case where two people managed to overwrite the work of each other.

zachberry:
Mockup - This shows an editor preview, however we talked and decided that for now that would be too complicated. So instead of an editor preview this would simply be a viewer preview. Then, the Note and 'View editor in new tab' items would be removed:

![image](https://user-images.githubusercontent.com/73479/74455619-96f34980-4e53-11ea-84f4-d54a8f8db6b4.png)

PR #1210, opened 2020-03-23
Module Restoration

ctcuff:
Adds ability to restore a module from a previous version 🎉. This isn't quite complete since some cleaning up can be done (and there are no tests). I'm submitting this now to open it up for feedback and to make sure this isn't going downhill. Here's the rundown:

Current look of the dialog (subject to change but also open to suggestions as well):
Screen Shot 2020-04-02 at 3 08 53 PM

**What needs work:**

- The SQL queries in `obojobo-repository/server/models/draft_summary.js` might not be 100% correct. I've only tested this with a single user so I'm not sure how it'd handle multiple users editing one module (it's probably gonna break?)

- The constructor for `draft_summary.js` was modified to include some entries from the `drafts_content` table in the DB. This works, but maybe it'd be better to make a separate class just for revisions.

- ~`obojobo-repository/shared/components/restore-module-dialog.jsx` uses `window.confirm` for confirmation before you restore a draft. I think using a modal would look cleaner, but `window.confirm` works too, especially since it's already being used when a draft is deleted.~ **(This has been added)**

- `obojobo-document-engine/src/scripts/viewer/util/api-util.js` and `obojobo-repository/shared/api-util.js` contain duplicated routes since I wasn't able to import those routes in `restore-module-dialog.jsx`.

- Because I added a "Restore" button to the "Module Options" dialog, on mobile, the top and bottom extend pass the screen.

- The restoration dialog doesn't look very good on smaller screens since I haven't added any media queries. I'm open to some design ideas for this but I'm honestly not sure how this should look. Maybe this is something we can revisit at a later date with a separate issue?

**What  ~hopefully~ actually works:**

- Anytime you save, you can look at a history of what you wrote and make that history the latest version (only tested with one user).

- Testing this on my end, there doesn't seem to be any conflict when you're editing a module and decide to restore a previous version. If you refresh without saving, the module you restored becomes the latest version. If you save your changes _after_ you restored a module, the module you saved becomes the latest version.

**What I'd like to add:**
_Note that most of the features below could be added at a later version or made into separate issues_

- ~When you restore a module, the scroll position of the revision history menu doesn't change. I think it'd be nice if the menu smooth-scrolled to the top to signify that the module you restored is actually the latest version.~ **(This has been added)**

- Possibly a confirmation that restoration was successful? 

- It'd be cool if the module title/dashboard updated if a user restored a module that previously had its title changed. If it's not too complicated, maybe we can achieve this with events.

- It'd be nice to know what date a draft was restored on. Currently, this would require either adding a column to the `drafts` or `drafts_content` table. Zach proposed a better idea of adding a `drafts_content_metadata` table. This could store what date a draft was restored on and the username of the person who restored it.

Resolves: #1142


packages/app/obojobo-repository/server/models/draft_summary.js
```diff
@@ -90,6 +97,49 @@ class DraftSummary {
...
+				repository_map_user_to_draft.user_id,
+				users.first_name,
+				users.last_name
+			FROM drafts_content
+			JOIN repository_map_user_to_draft
+				ON repository_map_user_to_draft.draft_id=drafts_content.draft_id
+			JOIN users
+				ON repository_map_user_to_draft.user_id=users.id
+			WHERE drafts_content.draft_id=$[draftId]
+			ORDER BY drafts_content.created_at DESC;
```
zachberry, line 116: Summoning @iturgeon

iturgeon: Yea, this is an awful lot of data and unbound by any restrictions.  While it may work for now, we need this to be paginated or not return `drafts_content.content`.

iturgeon:
Late to the party, but I kinda prefer restore.  Revert is much less used in English, making restore more understandable.  I even think restore makes more sense describing saving and old version on top of the newest version.

zachberry:
@iturgeon That's valid, restore is fine. I revisited my arguments from earlier and I'm not sure I agree with past me... not sure if there really is that much of a difference between the words on a practical level.

iturgeon:
Ok, I'm going to consider this done now :)

PR #1224, opened 2020-03-26
Filter out the Solution attr when converting Question JSON -> XML

### LINKED ISSUE #1222: Question conversion to XML doesn't filter out "solution"
A bug that was added by #1211

zachberry:
Fixes #1222


iturgeon:
We should add a test for the effects of this

iturgeon:
I added the test and tried it with the code before this pr (and the one it's spun out of). should cover our butts :)

PR #1242, opened 2020-04-03
MoreInfo (...) Button for MCChoice

### LINKED ISSUE #1212: Missing the (...) details menu for MCChoice elements
There's no way to get at these details so you can't set the ID of these without dropping to XML

vutoan1245:
Screen Shot 2020-04-03 at 2 14 29 PM

Top Right

Nodes that are not rendered in `obojobo-document-engine` will have to add MoreInfo Button(`...`) individually.

Resolve #1212

### DIFF
--- a/packages/obonode/obojobo-chunks-multiple-choice-assessment/MCChoice/editor-component.js
+++ b/packages/obonode/obojobo-chunks-multiple-choice-assessment/MCChoice/editor-component.js
@@ -3,12 +3,14 @@ import './editor-component.scss'
 
 import React from 'react'
 import Common from 'obojobo-document-engine/src/scripts/common'
+import MoreInfoBox from 'obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box'
 import withSlateWrapper from 'obojobo-document-engine/src/scripts/oboeditor/components/node/with-slate-wrapper'
 import { Transforms } from 'slate'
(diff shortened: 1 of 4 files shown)


packages/obonode/obojobo-chunks-multiple-choice-assessment/MCChoice/editor-component.js
```diff
@@ -65,10 +68,51 @@ class MCChoice extends React.Component {
 		)
 	}
 
+	duplicateNode() {
```
zachberry, line 71: This code seems like it's copied from oboeditor/components/node - Would it be possible to use those methods somehow? Or pull these functions into a util class which both this and Node could use?

vutoan1245: One possible solution is to move `duplicateNode` to the MoreInfoBox. We only need to pass in `element`. What do you think? 
We can do the same for a few other functions.

zachberry: That could work, I think I'd prefer to keep it in the base node class, but since we're only using it here I'd be fine with moving it.

vutoan1245:
I close this PR because it replicates the code from editor-component.js. As a result, any change in the editor-component.js will not refect here. There is a need for a better solution.

Issue #1225, opened 2020-03-27
Don't allow users to duplicate the Assessment section in the Visual Editor


zachberry:
Since we don't fully support multiple assessments we should disable the duplicate option for the Assessment.

PR #1243, opened 2020-04-03
Disable duplicate assessment

vutoan1245:
Resolve: #1225


keeeeeegan:
This looks good, however, I noticed this: (#1255) as a result of investigating this.

PR #1268, opened 2020-04-17
Keyboard Navigation

### LINKED ISSUE #691: Make editor navigation keyboard accessible
The editor navigation is only partially accessible by keyboard.  The following still need to be fixed:

- [x] Skip Navigation button/ Skip to Editor button
- [x] Quick navigation between nodes
- [x] Quick navigation to start and end of the page
- [x] Keystrokes to swap between node types
- [x] Titles for each button that list the keyboard shortcuts (not helpful for people with 0 mouse ability, but informative for power users and people with some-but-not-extensive mouse)
- [x] Keyboard Shortcuts listed on Menu buttons
- [x] Esc blurs editor, to prevent tab capture issue
- [x] Key to open More Info Box
- [x] Key to open insert above/below menu (CTRLSHIFT+ and CTRLSHIFT- ?)
- [x] Save works everywhere in doc

qwertynerd97:
Ensures that users without mouse access can still fully use the Visual Editor.

Resolves #691


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.js
```diff
@@ -31,6 +31,7 @@ class MoreInfoBox extends React.Component {
 		}
```
zachberry, line 31: General comment: Clicking on two (...) menus in a row results in the second focus not working

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.js
```diff
@@ -48,13 +49,26 @@ class MoreInfoBox extends React.Component {
 		this.close()
 	}
 
+	componentDidUpdate() {
+		// This autofocuses the id input box when the node is opened
+		// for easy use by keyboard users
+		if (this.state.isOpen || this.props.open) {
+			document.addEventListener('mousedown', this.handleClick, false)
+			this.idInput.focus()
```
zachberry, line 57: I think we should also `setTimeout(() => this.idInput.select(), 1)`

zachberry: I think we need to only do this when prevState.isOpen !== this.state.isOpen. When trying to (for example) change a page title the focus keeps returning to the id input

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/components/insert-menu.js
```diff
@@ -68,7 +68,7 @@ class InsertMenu extends React.PureComponent {
 
 			// Move left/up through the insert menu
 			// Left is logical for sighted keyboard users; up is a standard
-			// navigation principle for submenus
+			// navigation principle for submenus, and therefore logical to screenreaders
 			case 'ArrowLeft':
 			case 'ArrowUp':
```
zachberry, line 73: This doesn't seem to be working for me - If I'm in file and moving up and down the list, then press right, nothing happens. Unless this behavior is intended, but my expectation was for it to move to the next menu fwiw.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/page-editor.js
```diff
@@ -173,6 +179,91 @@ class PageEditor extends React.Component {
...
+
+			// Change the node so that the top insert menu is open
+			Transforms.setNodes(
+				this.editor,
+				{ open: 'bottom' },
+				{
+					at: nodeEntry[1]
+				}
+			)
+		}
```
zachberry, line 235: Cmd and + as well as Cmd and shift and + works. However only cmd and - works. Cmd and shift and - does not work.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/page-editor.js
```diff
@@ -173,6 +179,91 @@ class PageEditor extends React.Component {
...
+
+		// Open top insert menu
+		if (event.key === 'i' && (event.ctrlKey || event.metaKey) && event.shiftKey) {
```
zachberry, line 238: I can't get this to work because I think when shift is held down event.key is `I`, not `i`.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/format-menu.js
```diff
@@ -33,45 +35,55 @@ const FormatMenu = props => {
...
 			},
-			{ 
-				name: 'Code', 
-				type: 'action', 
-				action: () => props.editor.changeToType(CODE_NODE)  
+			{
+				name: 'Code',
+				type: 'action',
+				shortcut: 'Ctrl+Shift+C',
+				shortcutMac: '⌘+Shift+C',
```
zachberry, line 85: On the mac both cmd+shift+1 and ctrl+shift+1 work, however for this one only ctrl works. Is there an existing keyboard shortcut we're working around for this one?

packages/app/obojobo-document-engine/src/scripts/oboeditor/overwrite-bug-fixes.js
```diff
@@ -131,4 +131,29 @@ Transforms.mergeNodes = (editor, options) => Editor.withoutNormalizing(editor, (
 	if (emptyRef) {
 		emptyRef.unref()
 	}
-})
\ No newline at end of file
+})
+
+// This one is directly copied from Slate's master branch, but it has not been released yet
```
zachberry, line 136: I'm thinking we may want to throw an error if Transforms.setPoint is already set. That way in the future when we upgrade we'll remember this bit of code and be able to delete it.

packages/obonode/obojobo-chunks-list/converter.js
```diff
@@ -225,20 +229,91 @@ const switchType = {
...
+				// + the original offset
+				Transforms.setPoint(
+					editor,
+					{
+						path: lastLine[1].concat(relativeLeafPath),
+						offset: end.offset
+					},
+					{ edge: 'focus' }
+				)
+			}
```
zachberry, line 308: This code looks familiar, if multiple nodes are using it I'd like to move it into a reusable function (or if we could move some repeatable part out of it if they are indeed different)

zachberry:
One additional request - this was tested on dev/10 and not this branch, but it seems reasonable to put it here. Cmd/Ctrl+Shift+Z is an alternate for Cmd/Ctrl+Y in a lot of applications. If we're not already using it I'd like to support both for undoing. If this branch already has that in here then pretend I didn't say anything

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/list-dropper.js
```diff
@@ -93,56 +93,80 @@ class ListDropper extends React.Component {
...
+		const isMac = navigator.appVersion.indexOf('Mac') !== -1
+		// Decide whether or not to use the mac shortcut
+		// Note - users can spoof their appVersion, but anyone who is tech-savvy enough
+		// to do that is probably tech-savvy enough to know whether they use CTRL or ⌘
+		// for keyboard shortcuts
+		const shortcutMac = isMac && this.props.shortcutMac ? '\n' + this.props.shortcutMac : ''
+		// If the Mac shortcut exists, use it
+		// If there is no Mac shortcut and no mark.shortcut, the mac shortcut will be
+		// the blank string, so just use it
+		const shortcut = shortcutMac || !this.props.shortcut ? shortcutMac : '\n' + this.props.shortcut
```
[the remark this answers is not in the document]
qwertynerd97: The only place where we'd have a problem with this is in Format > Paragraph styles; I've set the Mac shortcuts to CTRL+SHIFT+1...6, because Cmd+Shift+3..5 are screenshot shortcuts.  But everything else has been changed to modifier + keys

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/components/insert-menu.js
qwertynerd97: In the File at the top of the editor, only the up/down works, since it's a vertically displayed menu (I couldn't figure out how to get the left/right working except as open/close because of the way the components are rendered).  In the inline insert menus, sighted users might expect to be able to use the left/right arrows since the menu is horizontal, so both the left/right and up/down scroll through the menu, and Escape closes the menu instead

packages/app/obojobo-document-engine/src/scripts/oboeditor/overwrite-bug-fixes.js
qwertynerd97: Transforms.setPoint is already set, but the current release has a bug where it doesn't alter the selection properly.  This code is taken straight from the master branch to fix that but has not been released in a version.

zachberry: Ah okay, I jumped to a conclusion that Transforms.setPoint was new. Is there a github issue or page that we could then refer to in the comments? Something so that later on when re-reviewing this code we'd be able to jump to that linked page and see if its been fixed.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/page-editor.js
qwertynerd97: I added the shifted versions of the keys, which should make shift work again.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/format-menu.js
qwertynerd97: I added the shifted versions of the keys here, which should make CMD work - it always worked on my machine, though, so let me know if it doesn't work

packages/app/obojobo-document-engine/src/scripts/oboeditor/util/selection-util.js
```diff
@@ -0,0 +1,49 @@
...
+	 * 
+	 * @param {Object} editor A Slate editor
+	 * @param {Array} path The path to the Obojobo node
+	 * @param {Object} point The original point that needs to be reset
+	 * @param {Array} leafPath The path to the leaf-most Element that contained the point (a child of path before conversion)
+	 * @param {String} leafSubtype The subtype of the leaf-most node that the new point will be inside
+	 * @param {start | anchor | end | focus} edge Which selection point is being set
+	 */
+
+	resetPointAtUncertianDepth(editor, path, point, leafPath, leafSubtype, edge) {
```
zachberry, line 17:
```suggestion
	resetPointAtUncertainDepth(editor, path, point, leafPath, leafSubtype, edge) {
```

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/drop-down-menu.js
```diff
@@ -81,6 +81,33 @@ class DropDownMenu extends React.PureComponent {
 		this.setState(prevState => ({ isOpen: !prevState.isOpen }))
 	}
 
+	renderAction(index, item) {
+		const isMac = navigator.appVersion.indexOf('Mac') !== -1
```
zachberry, line 85:
```suggestion
		const isMac = navigator.platform.indexOf('Mac') !== -1
```

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/editor-component.js
```diff
@@ -73,9 +86,59 @@ class Node extends React.Component {
...
+		)
+
+		// Handle cases where no nodes are selected because the page is changing
+		if (nodes.length === 0) {
+			return this.props.editor.toggleEditable(true)
+		}
+
+		// If the first or last node's open attribute matches the menu we are
+		// blurring, we need to clear the open attribute
+		if(menu === nodes[0][0].open) {
```
zachberry, line 108: Do you mind explaining this a bit, just trying to wrap my head around it. Am I right that each slate node in the editor has an `open` attribute and that dictates if the More Info Box is open? What does `[0][0]` refer to?

qwertynerd97: Yes, each node has an open attribute that can be either 'top', 'bottom', 'info', or null.  If it's not null, then the corresponding menu is open (top insert, bottom insert, or more info)
The Editor.nodes function returns a generator of NodeEntries, which is basically an array of [node, path] for each node it finds.  Therefore, nodes[0] is the first NodeEntry in the list, and nodes[0][0] is the actual slate node (nodes[0][1] is the path to the first returned node, which is used in the Transform)

qwertynerd97:
Modals seem to be having some extensive issues with tabbing, so that may need to be broken out into a separate issue, for a deep dive

PR #1280, opened 2020-04-22
Update ActionButton / MoreInfoBox to handle missing onClick Trigger

### LINKED ISSUE #1277: Editing triggers can crash the editor
I haven't dug into the cause yet, but adding a 'Hidden' trigger which 'Closes navigation' causes the page to crash. This may be an issue with multiple combinations of triggers/actions:

zachberry:
Fixes #1277


zachberry:
To test this: On current dev/10 click on an ActionButton to edit its triggers, then delete the onClick trigger and close the window. It should crash. On this branch it should handle that case.

There is an update to MoreInfoBox which updates its state when props.content changes - that's the change I'm most concerned about breaking something. However this solves this issue: If you click on the ActionButton and change triggers through that, then click on the MoreInfoBox to edit triggers that way they won't be updated until you click off to another node.

PR #1283, opened 2020-04-24
[wip] pulling list bullet change options back a bit

iturgeon:
Moving list bullet changing into the toolbar meant it
got much more complicated: dealing with converting and un-converting
selections as well as changing bullets at certain depths.

This reverts back to using inline buttons in the editor that
just alternates back and forth.  I left the drop down editor buttons
in the app, they are just display:none for now.

I also updated the list node's normalize script to use the list's
listStyle's as the authority of what bullets should be at each depth.

This also prevents hangingIndent from being added to text nodes in lists
if the option isn't already set.


iturgeon:
I'll need your eyes @qwertynerd97 I haven't updated tests or anything.  Doing my best not to loop too much into this and just focus on the task at hand - wrangling the new list features back a little bit to make them more predictable and buy some time to implement the full list toolbar options.

zachberry:
I made a few changes, but I'm running into failing tests that I'm hitting a brick wall in trying to fix. There seems to be an issue where lists won't normalize their type (as in a 'ordered' child item in a 'unordered' list should be converted to 'unordered' based on the intent of the test, but that doesn't seem to be happening and I'm not sure why)

PR #1284, opened 2020-04-27
[Fix] saving issue while editing MathEquation

### LINKED ISSUE #1278: Saving module while edit window open for Math Equation doesn't save new changes
To replicate:

1. Create a new Math Equation node
2. Click edit, type in any equation
3. Go to File > Save in the menu (without clicking on the document)
4. Click 'Preview Module'

The result is your equation won't appear in the preview since the equation you typed hasn't been "committed" to the document.

vutoan1245:
Resolve #1278

### DIFF
--- a/packages/obonode/obojobo-chunks-math-equation/editor-component.js
+++ b/packages/obonode/obojobo-chunks-math-equation/editor-component.js
@@ -26,20 +26,24 @@ class MathEquation extends React.Component {
 	constructor(props) {
 		super(props)
 
+		// copy the attributes we want into state
+		this.state = this.contentToStateObj(this.props.element.content)
+
+		// used to reduce the speed/cost of changes so typing isn't slow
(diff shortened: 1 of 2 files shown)


vutoan1245:
@qwertynerd97 I updated this PR to add saving whenever there are changes. I feel like this will lead to fewer unintended errors and more consistent code. What do you think?

packages/obonode/obojobo-chunks-math-equation/editor-component.js
```diff
@@ -36,6 +36,7 @@ class MathEquation extends React.Component {
 		const content = this.props.element.content
 		this.state = this.contentToStateObj(content)
 
+		this.path = ReactEditor.findPath(this.props.editor, this.props.element)
```
iturgeon, line 39: Caching the path in the constructor may be a little dangerous.  If this component is reused but moved around in the document - It's path will change but not get updated here.

PR #1289, opened 2020-04-29
Fixed errors in input on question banks

qwertynerd97:
Fixes an error with the input box on question banks, where the cursor's focus could jump around while changing the number of questions to select.


iturgeon:
This'll need a pass over by @qwertynerd97 again. Turns out dev/11's qb editor component has been changed into a functional component.  So the changes you added may not be needed / present different problems now.

However, you and I ended up adding some other fixes in (style, test changes) that we def want to keep (I think).  So I did my best to consolidate the differences.

This is quite different from the original pull request at this point.

Questions remaining:

* Does adding freeze/unfreeze to the radio buttons still make sense?
* Do the flex box style changes still apply to the functional component?

qwertynerd97:
It looks like the fix for the input reverted after the merge, although the formatting changes are still an issue.  I'll poke at it and see if I can fix the regression - it may need to move back to a class component instead of a functional

iturgeon:
Hmm, things do look strange here (not seeing your commits 🤷 )  I revived the original over here for comparison https://github.com/ucfopen/Obojobo/pull/1345

PR #1297, opened 2020-05-07
Prevent module titles from being set to an empty string.

### LINKED ISSUE #1037: Don't allow modules with blank titles
In the visual editor you can have a blank title - if you do this the UI gets upset since the title string is now an empty element. We don't really want modules to have blank titles anyway so this should be prevented in the editor. If someone creates a blank title in XML and views it in the visual editor then at least the UI should display the empty string in a way that doesn't mess up the page.

FrenjaminBanklin:
Resolves #1037.

The module title adjustment dialog in the Visual editor's nav area will now prevent saving and display an error if the 'Module Title' input field is blank when the Done button is clicked.

The editor title input component is now a div containing an input and an optional div rather than a single input - if the input is empty and is then blurred, rather than saving the module title it will re-focus the input and render the adjacent div with a warning message.


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/editor-title-input.js
```diff
@@ -7,47 +7,72 @@ const EditorTitleInput = ({ renameModule, title }) => {
...
+				placeholder="Module Title"
+				onChange={event => {
+					if (emptyTitleError && event.target.value !== '') setEmptyTitleError(false)
+					setStateTitle(event.target.value)
+				}}
+				onBlur={event => {
+					if (stateTitle !== title) {
+						const newTitle = stateTitle.trim()
+
+						if (newTitle !== '') renameModule(newTitle)
```
zachberry, line 39: This is a nitpick, but prefer ifs to use `{ }` if used with an else.

PR #1308, opened 2020-05-07
adds xml snapshots to json documents so they are easier to read in diffs

iturgeon:
This should be added to dev/11 when it's ready


iturgeon:
I'm just going to cancel this for now.

PR #1310, opened 2020-05-08
Make keys for ActionButtonEditorAction components more specific to avoid collisions for duplicate actions.

### LINKED ISSUE #1180: Triggers dialog can have duplicate keys
Related #1169 

To replicate:

1. Create an action button
2. Edit triggers to bring up the triggers dialog
3. Create two duplicate actions (such as 'Start an attempt for')
4. Close the dialog

Result:

```
Warning: Encountered two children with the same key, `assessment:startAttempt`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

FrenjaminBanklin:
Fixes #1180.

Appends the index of a click trigger's action to the action type when generating a key for the React component to prevent errors due to duplicate keys.

### DIFF
--- a/packages/obonode/obojobo-chunks-action-button/editor-component.js
+++ b/packages/obonode/obojobo-chunks-action-button/editor-component.js
@@ -55,8 +55,8 @@ class ActionButton extends React.Component {
 					<div className="trigger-list">
 						<div className="title">When the button is clicked:</div>
 						{isAnOnClickActionSet ? (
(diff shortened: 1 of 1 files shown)


iturgeon:
Tested, works!  I realized, however, that the key is more complex than it needs to be, so I simplified it.

PR #1326, opened 2020-05-19
Keep styles buttons highlight when its corresponding text is selected

vutoan1245:
Example for Bold, Italic, Strikethrough: 

Style-buttons-highlight-feature


vutoan1245:
@zachberry 
About point 1: I was using Slate's native function, so it is up to Slate to decide on it. 
I cannot find a solution for it.

qwertynerd97:
Slate decides marks based on the location of the start of the selection, rather than the whole selection.  I think it makes sense to keep the highlighting in the toolbar as it is currently because it reflects the behavior that will occur when pressing it - highlighted buttons will remove the mark and un-highlighted will add the mark.  Since we are using the same Slate function to determine highlighting and toggling, the behavior is at least internally consistent.

zachberry:
@qwertynerd97 Thanks for the explanation, that makes sense. In that case, ignore point 1.

iturgeon:
just to be clear and ping @vutoan1245 can you look at what it would take to get the link button to behave the same?

vutoan1245:
@iturgeon 
I am looking into it. I still haven't figured out why the link-mark behaves differently compared to other marks.
I managed to make to partially work by calling `editor.toggleMark(BOLD_MARK)` under `action` in `link-mark.js`. It stops working after refreshing the page.

vutoan1245:
The build should work now

zachberry (dismissed):

vutoan1245:
The changes look good.

iturgeon (approved):
Looks good.  Nice use of `test.each` @vutoan1245

PR #1328, opened 2020-05-20
Optimizations and tweaks to inline latex

zachberry:
* HoveringPreview now uses display none when not being shown instead of opacity=0, additionally doesn't render if there is no latex to generate.
* HoveringPreview now has a max-width of 60vw to prevent malformed latex from growing the preview off the screen
* Slightly narrows the inline latex font in the editor so it takes up less space
* Disables spellcheck in inline latex in the editor


iturgeon:
I really messed that one up - force pushed the base branch to this PR which caused github to close it (no changes to merge in) and now that it's closed I can't force push to fix it :/  

I'm reopening over here #1339

PR #1329, opened 2020-05-20
Fixes Editor Done button

### LINKED ISSUE #1327: Clicking done after editing a page title doesn't do anything sometimes
Looks like a bug introduced in v9.0.0.  Sometimes if you edit a page title or id, the done button doesn't do anything

zachberry:
Resolves #1327


iturgeon:
FWIW I ended up overwriting Zach's fix by removing the offending code all together.  

As shown in the editor-nav, `sub-link` items aren't used

https://github.com/ucfopen/Obojobo/blob/master/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/editor-nav.js#L93-L122

PR #1332, opened 2020-05-22
Speedup visual editor

iturgeon:
* `.bind()` caching implemented across the app
* FileToolBar re-organized to reduce redraws and work on redraws
* Removes unused Table GridButton


iturgeon:
I noticed a bug w/ the move up/down buttons because zach mentioned it - moving an item down twice works, then the third time it just keeps swapping back and forth.

PR #1335, opened 2020-05-22
allow fractional MathEquation sizes, min/max input

### LINKED ISSUE #1333: Math Equation size isn't rendered in veiwer
Setting a size of .5 doesn't reduce the size in the viewer.  

It looks like the adapter is using parseInt where it should use parseFloat or just pass the string through?

iturgeon:
Some math equations are too big to be displayed in the viewer.
This will allow float values for size in the viewer.

In addition, the visual editor input for size now has a min
and max value (0.1 - 20).  Both are beyond advisable limits, but
provided for the occasional edge case.

fixes #1333


zachberry:
I made a few more changes - 

There were two cases where the editor still didn't match the viewer:

* Can set size to zero in the editor, resulting in no equation shown in the editor but you can view it in the viewer
* Can set the size to some value other than one, then delete the value in the editor and the equation in the editor won't change, but it will be set to the default in the viewer

Additionally, it was possible to set values outside of 0.1 to 20 in the XML/JSON, so the viewer now also restricts values to this range.

Issue #1334, opened 2020-05-22
Sometimes a MathEquation won't save right after editing


iturgeon:
1. Add a Page
2. add an equation
3. set value to `W_{23}=P_{avg}\left[V_3-V_2\right]=\left(\frac{P_3+P_2}{2}\right)\left(V_3-V_2\right)=\left[\left(\frac{3+1}{2}\right)kPa\right]\left(1-5\right)m^3\left|\frac{10^3N/m^2}{1kPa}\right|\left|\frac{1kJ}{10^3N\cdot m}\right|=-8kJ`
4. Preview
5. Add another equation after 
6. Preview again

iturgeon:
duplicate of #1278

PR #1336, opened 2020-05-23
mathEquation editor saves more often

iturgeon:
MathEquation would previously only save when the edit dialog
switched from open to closed.  Unfortunately it doesn't close
when you click anywhere that's not in the slate display.

Ex: edit contents of the equation and click save or preview from the
toolbar.  The contents aren't copied to the slate document and when
save is called, the changes don't make it into the document

This adds a debounced function to make sure the document is saved
as the content changes.

It also second condition in componentDidUpdate that'll cause the
contents to be copied to slate - when the editor is closing (previously
was just when it lost selection)

At the same time I moved the code that closes the editor out
of the function intended to push updates to slate.

solves #1334 #1278

### DIFF
--- a/packages/obonode/obojobo-chunks-math-equation/editor-component.js
+++ b/packages/obonode/obojobo-chunks-math-equation/editor-component.js
@@ -26,19 +26,24 @@ class MathEquation extends React.Component {
 	constructor(props) {
 		super(props)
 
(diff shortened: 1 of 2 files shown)


iturgeon:
contains duplicate work from #1284 so closing and contributing to that pr instead

PR #1368, opened 2020-06-01
XML editor crashes with visual editor tutorial

### LINKED ISSUE #1363: 9.1.0 Review: Attempting to edit a new Visual Editor tutorial in XML crashes
It results in a gateway timeout:

image

vutoan1245:
Resolve #1363 

Reason: json-to-xml parser crashes because `window.katex.renderToString` is undefined in nodejs
`nodeToWrap.html` is not necessary for the parser so I bypass it using try and catch

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/common/text/styleable-text-renderer.js
+++ b/packages/app/obojobo-document-engine/src/scripts/common/text/styleable-text-renderer.js
@@ -159,8 +159,13 @@ const wrapElement = function(styleRange, nodeToWrap, text) {
 			)
 			nodeToWrap.parent.replaceChild(nodeToWrap, newChild)
 			newChild.addChild(nodeToWrap)
-			const html = window.katex.renderToString(text, { throwOnError: false })
-			nodeToWrap.html = `<span aria-hidden="true">${html}</span>`
+
+			try {
+				const html = window.katex.renderToString(text, { throwOnError: false })
+				nodeToWrap.html = `<span aria-hidden="true">${html}</span>`
+			} catch (e) {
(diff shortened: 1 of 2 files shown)


packages/app/obojobo-document-engine/src/scripts/common/text/styleable-text-renderer.js
```diff
@@ -159,8 +159,13 @@ const wrapElement = function(styleRange, nodeToWrap, text) {
...
 			newChild.addChild(nodeToWrap)
-			const html = window.katex.renderToString(text, { throwOnError: false })
-			nodeToWrap.html = `<span aria-hidden="true">${html}</span>`
+
+			try {
+				const html = window.katex.renderToString(text, { throwOnError: false })
+				nodeToWrap.html = `<span aria-hidden="true">${html}</span>`
+			} catch (e) {
+				console.error(e) // eslint-disable-line
+			}
```
iturgeon, line 168:
```suggestion
			// if we're running in the browser, render katex
			// note: this does get called on the server, but we don't need the rendered katex there
			if(typeof window !== "undefined"){
				const html = window.katex.renderToString(text, { throwOnError: false })
				nodeToWrap.html = `<span aria-hidden="true">${html}</span>`
			}
```

Some observations: 

- prefer not to console.error here, especially when it's expected to throw errors for the server (keep logs lighter, free of miss-reported errors)
- this would be more useful as an if - it'll self document a little better (if isBrowser do ...)

vutoan1245: I like this implementation

PR #1372, opened 2020-06-02
[fix] repository links to classic editor

### LINKED ISSUE #1341: Need to remove `classic` editor return from buildQueryWhere
Removal of the classic route (https://github.com/ucfopen/Obojobo/commit/f5c7d8a8eecaa1725381f815f1bbd249ad98bd52) broke the dashboard for me - see:

It appears that this "classic" value is still left in via the buildQueryWhere function: 

https://github.com/ucfopen/Obojobo/blob/master/packages/app/obojobo-repository/server/models/draft_summary.js#L17

I believe this CASE statement is no longer necessary since we always put them into the visual editor.

iturgeon:
some parts of the classic editor links were removed, but others remained.  It's long gone now, so it needed to be removed from the repository.

This should be every remaining part of the 'classic' mode editor urls.

fixes #1341


iturgeon:
I'm going to merge this now since it's pretty low risk w/o extra revies

PR #1378, opened 2020-06-03
adds close button to modals and allows escape to close

### LINKED ISSUE #1377: using escape to exit iframe (and other chunks) leaves editor disabled

iturgeon:
closes #1377


iturgeon (commented):
Tested in browser, tried all the modals I could find.

* esc closed them
* the x closed them
* the x didn't sit on top of the content
* didn't see any x's on dialogs that shouldn't be able to be closed
*

zachberry (changes requested):
I'm reviewing both this and #1723 which are very similar. Both though show a (X) button on modals where `preventEsc` is true - for example, the Resume Attempt dialog. That means you can click (X) and you won't have to resume the attempt.

Additionally, I'm not sure some of the modals need the (X) to show up. For example, viewer alerts like in the screenshot posted in my earlier review.

Here's my thought

* If modal has a `onClose` prop method it shows the (X) and allows ESC. If it doesn't, there is no (X) and ESC is disabled, and it's up to the modal to make sure there's a way to close it.
* An optional param, `showCloseButton` can be set to `false`. That means if there should be a close button (based on the existence of an `onClose` prop), setting this to false will hide it. Needed to simplify small dialogs. ESC would still work.
* Remove `preventEsc`. Anywhere we're passing `preventEsc`, just make sure we're not passing an `onClose` method, and ESC will be prevented.

PR #1382, opened 2020-06-08
Add saving indicator

### LINKED ISSUE #1343: Editor sometimes displays "Saved" when not saved
Found via #1342 

Want to make sure the 'Saved' text only comes up when the save API call returns successfully.

vutoan1245:
Resolves #1343 

"Saved" indicator only appears when API call returns successfully

Also added "Saving..." indicator:
Screen Shot 2020-06-08 at 12 55 50 PM


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/visual-editor.js
```diff
@@ -299,7 +302,7 @@ class VisualEditor extends React.Component {
 		// This mostly happens with MoreInfoBoxes and void nodes
 		if (this.editor.selection) this.editor.prevSelection = this.editor.selection
 
-		this.setState({ value, saved: false })
+		this.setState({ value, saved: false, saving: false })
```
zachberry, line 305: This causes issues when, for example, when `saving` = `true` but the user clicks somewhere on the document. Then onChange is fired and the 'Saving...' text goes away.

I think instead we should coalesce this into a single property with multiple states (since `saved` and `saving` could theoretically both be true but that's an invalid state). My first idea is that we could change this to `saveState` which could be set to `"idle"`, `"saving"`, `"save-successful"` and `"save-failed"` - but I'm open to discussion.

vutoan1245: A click is registered as an edit, so I think it makes sense to stop showing `Saving...`

I change the state to `saveState`. I agree that it is a lot more flexible.

zachberry (changes requested):
The last issue is that you can still have the 'Saving...' text disappear if you click somewhere on the page while the save request hasn't completed. As is it appears to the user that the saving is cancelled when clicking somewhere on the page.

Marking as 'request changes', but if we want to move that to a new issue I'm okay with merging this in now and enhancing it later.

zachberry (approved):

jpeterson976 (dismissed):
LGTM! I think the clicking thing can be made into its own issue, this seems to be in a good enough state to ship.

PR #1385, opened 2020-06-08
Restrict width more info box input

### LINKED ISSUE #1265: Restrict width MoreInfoBox inputs
Just found this while testing long page titles:

image

vutoan1245:
Resolves #1265


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.scss
```diff
@@ -105,6 +105,10 @@
 					}
 				}
 
+				select {
+					width: 22.7em;
```
zachberry, line 109: I think we should go with a flex box solution. This will break as soon as we need to change the dimensions of the more info box.

vutoan1245:
Bonus:
- input and select now have the same font and height
- label is connected with input/select
- "Copy Id" and "Edit" buttons now have the same width
- space before "Copy Id"

zachberry (approved):
LGTM!

zachberry (dismissed):
Forgot I was merging in dev/13, so here's a second review, but anyway, LGTM!

ctcuff (approved):
Looks great!

PR #1387, opened 2020-06-09
Keep the navigation bar stays on the left

### LINKED ISSUE #1263: Force nav bar in visual editor to stay at the top of the editing area
The nav bar in the viewer stays static at the top, so for similarity (and for ease of editing long documents) we should also fix this to the top.

I think enclosing the nav in its own box will let us achieve this, and I think it's acceptable in the context of the editor.

Mockup:

image

vutoan1245:
Resolves #1263 

Summary:
- Keep the navigation bar stay the left when scrolling
- Get rid of space around the main content


vutoan1245:
This solution does not work.

PR #1410, opened 2020-07-13
Easy way to move questions to questionbank

### LINKED ISSUE #1167: Easily be able to move questions in/out of QuestionBanks
Real world issue: Instructor created a bunch of questions but then realized they should have been in a QuestionBank - as of now the only way to fix this without re-authoring questions is to drop to XML.

There's also a need to take a practice question, copy it, then paste it inside a QuestionBank to then modify as an assessment question.

Related: #1125 

I'm picturing a menu on QuestionBanks to let you paste in a question you copied (since it's not possible to do this with the normal copy/paste inline). It would also be able to wrap selected questions in a question bank as well as un-wrap a question bank - Much like grouping/ungrouping in a graphic program.

vutoan1245:
Resolve #1167


vutoan1245:
Note: Because we cannot use the Viewer to render components in the Editor, I added a new file `model-render.js` that uses the viewer components and dynamically imports it when needed.
Not sure if this is the best solution.

packages/obonode/obojobo-chunks-iframe/viewer-component.js
```diff
@@ -38,7 +38,9 @@ export default class IFrame extends React.Component {
 		this.boundSkipToBottom = this.onClickSkipToBottom.bind(this)
 		this.boundSkipToTop = this.onClickSkipToTop.bind(this)
 
-		MediaUtil.setDefaultZoom(this.props.model.get('id'), this.props.model.modelState.initialZoom)
+		if (!props.readOnly) {
```
zachberry, line 41: What's the need for this addition?

packages/obonode/obojobo-chunks-question-bank/model-renderer.js
```diff
@@ -0,0 +1,87 @@
...
+			YoutubeAdapter.construct(model, model.attributes)
+			break
+		default:
+			return null
+	}
+
+	return <Component model={model} moduleData={{ isFocussed: false }} readOnly />
+}
+
+export default ModelRenderer
```
zachberry, line 87: This is problematic because it's another place to define code for chunks, should any more be added to the system, and probably easy to miss. Would it be possible to do some of this code in the Registry?

packages/obonode/obojobo-chunks-iframe/viewer-component.js
vutoan1245: Because `MediaUtil` is not initialized in the editor and we don't need options for zoom. Not really sure if that is a good solution.

packages/obonode/obojobo-chunks-question-bank/model-renderer.js
vutoan1245: - The current way to add it to the Registry is to import `viewer.js` of each chunk, but this will mess up the Registry of the editor and cause it to break.
- It is possible to add each component individually but we will have the same problem. My biggest concern is that it will significantly increase the package size.

packages/obonode/obojobo-chunks-question-bank/import-questions-modal.scss
```diff
@@ -0,0 +1,57 @@
...
+
+.import-questions-model {
+	max-width: 70vw;
+
+	.import-model--question-content {
+		background: $color-bg2;
+		border-radius: $dimension-rounded-radius;
+		max-height: 55vh;
+		min-width: 35em;
+		overflow-y: scroll;
```
ctcuff, line 11: Super minor, but I think this would look a little cleaner with `overflow: scroll` so we don't have an empty scrollbar when the dialog doesn't have a lot of content.

ctcuff (commented):
Overall I think it looks great! Tested on FF, Chrome, and Safari and it's looking good. One tiny thing I noticed is that you can only import questions after you've already saved a module, not really a major issue though.

zachberry (approved):
Awesome! LGTM.

jpeterson976 (approved):
LGTM! This is great, definitely a very useful feature to have.

PR #1411, opened 2020-07-14
Heroku deploy fixes

iturgeon:
Heroku deploy broke due to 2 reasons - the post deploy mechanism wasn't running the migration scripts and new ssl config options / requirments for bypassing self-signed certs on free heroku dbs - combined with the fact that obojobo config supports json env var expansion but db-migrate doesn't have these features.

* post-deploy processes moved to Procfile from app.json
* objobo-migrate upgraded to provide obojobo config abilities to db-migrate

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+release: yarn obojobo-migrate up -e production
+web: yarn start
--- a/app.json
+++ b/app.json
@@ -6,7 +6,6 @@
 	"logo": "https://ucfopen.github.io/Obojobo-Docs/assets/images/obojobo_blue.svg",
 	"success_url": "/",
 	"scripts": {
-		"release": "cd node_modules/obojobo-express && yarn db:migrateup",
 		"postdeploy": "cd node_modules/obojobo-express && yarn sampleDraft:seed",
 		"purgedata": "yarn run obojobo_purge_data"
 	},
--- a/packages/util/obojobo-lib-utils/bin/obojobo-migrate
+++ b/packages/util/obojobo-lib-utils/bin/obojobo-migrate
@@ -18,7 +18,7 @@ const fs = require('fs-extra')
 const os = require('os')
 const { execSync } = require('child_process')
 const dbMigratePath = require.resolve('db-migrate/bin/db-migrate')
-const configPath = require.resolve('obojobo-express/server/config/db.json')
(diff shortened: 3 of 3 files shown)


Procfile
```diff
@@ -0,0 +1,2 @@
+release: yarn obojobo-migrate up -e production
```
iturgeon, line 1: As it turns out, this is how release scripts are run.  I either made a mistake or the documentation changed from the last time I worked on this (see the `release` script added to app.json)

app.json
```diff
@@ -6,7 +6,6 @@
 	"logo": "https://ucfopen.github.io/Obojobo-Docs/assets/images/obojobo_blue.svg",
 	"success_url": "/",
 	"scripts": {
-		"release": "cd node_modules/obojobo-express && yarn db:migrateup",
```
iturgeon, line 9: I guess this doesn't belong here.

packages/util/obojobo-lib-utils/bin/obojobo-migrate
```diff
@@ -18,19 +18,23 @@ const fs = require('fs-extra')
...
 	const tempDir = fs.mkdtempSync(`${os.tmpdir()}/obojobo-migrations-`)
-
+	const configPath = `${tempDir}/dbconfig.json`
+	console.log(tempDir)
 	try {
 		// useful to understand how args are being inserted into the command
 		command = `${dbMigratePath} ${args} --config ${configPath} --migrations-dir ${tempDir}`
 
+		// use obojobo's config parser to create a json file for db-migrate
+		fs.writeFileSync(configPath, JSON.stringify({development: config.db, production: config.db}, null, '\t'))
```
iturgeon, line 36: This little change asks obojobo's normal config to process the config files, convert them back into json - and write them in a format that db-migrate will fall use.

Bonus: NODE_ENV will govern the config chosen when running the server and when running the migrations.

Bonus: Obojobo allows ENV variables ending in `_JSON` to be automatically expanded in db.json config.  db-migrate didn't handle those.  Making it easy to forget that you couldn't use json expansion in db.json in some situations.  This fixes that, so there should be no differences!

iturgeon:
I tested this locally and on Heroku.  It looks like it's working.  Also tested adding a test migration commit to see if Heroku's deploy would run the migration script.  It did!

PR #1433, opened 2020-07-21
.text className causes katex rendering issue

### LINKED ISSUE #1428: `.text` css class breaking latex which uses `\text` functions
We have a few cases in the css where we have a very basic `text` class that ends up being applied unexpectedly. That 'text' class should be removed or made specific such that it only styles what we expect.

Example: Make inline latex or a Math equation with `T(^{\circ}\text{R})=1.8(\text{K})`.

Expected result:

Screen Shot 2020-07-20 at 10 45 06 AM

Result in the viewer:

Screen Shot 2020-07-20 at 10 44 52 AM

vutoan1245:
Change from `.text` to .`text-group`

Resolves #1428


vutoan1245:
@zachberry agree

PR #1438, opened 2020-07-23
Issue/1234 youtube user friendly

### LINKED ISSUE #1234: Make time and ID options for YouTube node more user friendly
Right now, the YouTube Video dialog only supports adding a video through its ID. The start time and end time fields also only support seconds. A better approach would be to allow users to enter the entire video url (so something like `https://www.youtube.com/watch?v=dQw4w9WgXcQ` instead of `dQw4w9WgXcQ`). Same thing applies to the start and end time fields. We should allow times like `1:30` or even `1m30s`.

keeeeeegan:
Replaces the "video id" field with a url field, that allows users to paste in the youtube url, embed link, or embed code.

Resolves #1234


zachberry:
This should be ready for re-review - It's been updated with dev/16 and should support more URL formats

ctcuff (approved):
LGTM!

jpeterson976 (approved):
Works great, LGTM!

PR #1445, opened 2020-07-29
Added large option to iframe component

### LINKED ISSUE #1405: Give iframes the option to be at the larger width (same as questions, Figures set to large and YouTube videos)
You can see an example here:

image

The first element was done with a manual <iframe...> command, the second was done with our IFrame, the last is a YouTube video. I can see the desire for IFrames and YT videos to be set at the same width.

maufcost:
Added the option for the **size** of iframes to be large or medium. If they're large, they will extend to the same width as youtube iframes.
Fixes #1405


packages/obonode/obojobo-chunks-iframe/converter.js
```diff
@@ -23,13 +23,14 @@ const slateToObo = node => ({
 		height: node.content.height,
 		initialZoom: node.content.initialZoom,
 		autoload: node.content.autoload,
-		controls: node.content.controls
+		controls: node.content.controls,
+        size: node.content.size
```
jpeterson976, line 27: Pretty minor, but this line is indented using spaces while the rest of the file uses tabs.

jpeterson976 (changes requested):
Works great! The IFrames page in the Visual Editor Tutorial should be updated to include the new option, though.

PR #1450, opened 2020-08-05
adds viewer:initialView event and fixes opening in a new tab

### LINKED ISSUE #1449: Cmd click to launch in a new tab causes visibility event problems
If you use the dev dashboard to launch a student view in a new tab, it'll launch that tab without switching focus to it.

Obo will begin loading in that tab.

Once you switch to that tab, obo thinks you must have left and come back, so the code tries to send a return event - with a reference to the leave event.  But there was none.

This brings up the point that timing data could be a little less reliable as a result.  

So I suggest we add a new event that signals the real first view of the frame (previously we assumed when you started a visit, you were viewing), and handle the case that there is no leave event to reference on the first visit.

iturgeon:
Fixes #1449 

Problem: if you use the dev dashboard to launch a student view
in a new tab, it'll launch that tab without switching focus to it.

Once you switch to that tab, obo thinks you must have left and
come back, so the code tries to send a return event - with a
reference to a non-existent leave event.

This commit adds a new event to represent the first time the
module is visible to the student.  It will increase timing
accuracy and prevent the error thats currently being thrown
due to an undefined leave event reference.


packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -335,8 +339,13 @@ export default class ViewerApp extends React.Component {
...
 			})
-		} else {
-			APIUtil.postEvent({
+
+			return
+		}
+
+		// returning
+		if (this.leaveEvent) {
+			ViewerAPI.postEvent({
```
zachberry, line 348: ViewerAPI is not defined here

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -350,7 +359,21 @@ export default class ViewerApp extends React.Component {
 
 			delete this.leaveEvent
 			delete this.leftEpoch
+			return
 		}
+
+		// first view
+		this.sendViewStart()
```
zachberry, line 366: It's probably fine, but I'd feel a little bit better using a flag. Something either like `this.isViewed` or `this.state.isViewed`, maybe `this.hasSentInitialView`? - Then inside `sendViewStart` it could set the flag to true and then no longer allow the event to be sent once true.

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
iturgeon: bah, I stashed it off the branch I was working on - it will be :( Too many spinning plates

iturgeon: I this was finally resolved upstream as ViewerAPI was added in

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
iturgeon: I changed the name some, it still doesn't conform to this suggestion, but it's better than what it was.

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -341,22 +345,44 @@ export default class ViewerApp extends React.Component {
...
 			})
-		} else {
+
+			return
+		}
+
+		// From Hiding to Viewing
+		if (this.viewerHideDate) {
+			// leaveEvent may not exist if postEvent for 'viewer:leave' didn't complete
+			const relatedEventId = this.leaveEvent?.extensions?.internalEventId ?? 'not availible'
```
zachberry, line 355:
```suggestion
			const relatedEventId = this.leaveEvent?.extensions?.internalEventId ?? 'not available'
```

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -341,22 +345,44 @@ export default class ViewerApp extends React.Component {
...
+
 			ViewerAPI.postEvent({
 				draftId: this.state.model.get('draftId'),
 				action: 'viewer:return',
 				eventVersion: '2.0.0',
```
zachberry, line 360:
```suggestion
				eventVersion: '3.0.0',
```

My thought for this - `relatedEventId` used to expect null or some id, now it may report "not available". I'm open to discussion on that though.

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
zachberry: I'm not seeing something, but it seems to be that sendViewStart will get called every-time onVisibilityChange gets called (although that doesn't actually appear to be happening). Also thinking we should change to `sendInitialViewEvent` so it doesn't get confused with the similarly named visit:start event

zachberry (changes requested):

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
iturgeon: return statements are easy to miss above

zachberry (approved):
I added in the following:

* Fixes typo in “not available”
* Updates viewer:return to 3.0.0
* Renames sendViewStart to sendInitialViewEvent
* Updates breaking viewer-app tests

iturgeon (commented):
Looks good after @zachberry's changes

PR #1451, opened 2020-08-06
Issue/778 use intersection observer

### LINKED ISSUE #778: Use IntersectionObservers instead of getBoundingClientRect calls in scroll listener to determine what is visible on screen
Possibly related: #666 😈

More info: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API and https://developers.google.com/web/updates/2016/04/intersectionobserver

For example, the background-muted effect that happens when you view a question has a feature that returns the page back to normal when you scroll past the question. This is done in a a scroll event, calling our Screen.isElementVisible function which uses getBoundingClientRect. getBoundingClientRect forces a page layout so doing that on scroll is pretty rough. Sounds like IntersectionObservers solve this problem. We can fallback to the current method for IE11.

This may open the door to being able to send events about what's visible on the screen in an efficient way.

jpeterson976:
Replaced getBoundingClientRect calls and scroll listener with an IntersectionObserver.
Resolves #778


packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -96,7 +96,9 @@ export default class ViewerApp extends React.Component {
 		this.onVisibilityChange = this.onVisibilityChange.bind(this)
 		this.onMouseDown = this.onMouseDown.bind(this)
 		this.onFocus = this.onFocus.bind(this)
-		this.onScroll = this.onScroll.bind(this)
+		this.startObservingForIntersectionChanges = this.startObservingForIntersectionChanges.bind(this)
+		this.onIntersectionChange = this.onIntersectionChange.bind(this)
+		// this.onScroll = this.onScroll.bind(this)
```
zachberry, line 101: Make sure to remove this any any other commented out code

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -255,6 +257,7 @@ export default class ViewerApp extends React.Component {
 
 		// use Focus Store values to update DOM Focus
 		this.updateDOMFocus()
+		this.startObservingForIntersectionChanges()
```
zachberry, line 260: Need a parallel `stopObservering...`, this probably isn't too bad but I'm concerned that we're not disconnecting the observers. I don't know the reality but it seems possible to create memory leaks.

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -396,7 +399,7 @@ export default class ViewerApp extends React.Component {
 		}
 	}
 
-	onScroll() {
+	startObservingForIntersectionChanges() {
 		const focusState = this.state.focusState
```
zachberry, line 404: I think here we should go ahead and kill the existing observer. My thought is you could just call `this.stopObservingForIntersectionChanges()` which would be something like

```javascript
if(!this.observer) return;

this.observer.disconnect();
delete this.observer
```

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
```diff
@@ -412,11 +415,48 @@ export default class ViewerApp extends React.Component {
...
+	}
+
+	onIntersectionChange(changes) {
+		const change = changes[0]
+
+		if (change.intersectionRatio > 0) {
+			return false
 		}
+
+		FocusUtil.clearFadeEffect()
```
zachberry, line 435: For now we're only using the onScroll (and now, IntersectionObservers instead) to determine when to call this clearFadeEffect method, and once it's cleared it doesn't come back. So go ahead and kill the observer with `stopObservering...`

packages/app/obojobo-document-engine/src/scripts/viewer/components/viewer-app.js
zachberry: Now that you have the `stopObserving...` method it needs to be called here

zachberry (dismissed):
Tested in Chrome, FF, Safari, all looks good, code looks good.

ctcuff (dismissed):
LGTM!

PR #1459, opened 2020-08-11
Color Text

### LINKED ISSUE #1162: Colored text
Got a feature request to add colored text in Obo documents. I think we want to restrict the colors to maintain contrast for accessibility, but here's a use case (screenshot from another LMS platform):

image

image

image

vutoan1245:
Resolve #1162 

Screen Shot 2020-08-11 at 10 24 20 AM

I am opening for any suggestions/improvements


vutoan1245:
@zachberry It should address all mentioned issues except the default color options.

I fixed the selection color using this:
```
span::selection {
	background-color: highlight;
	color: inherit;
}
```
The background color does not affect the colored text but I am opening to make a custom color.

vutoan1245:
The current JSON/XML structure for color text:

```
"text": {
    "value": "Your text here",
    "styleList": [
        {
            "end": 9,
            "data": {
                "text": "rgb(255, 0, 0)"
            },
            "type": "color",
            "start": 5
        }
    ]
}
```
```
<t>Your <color text="rgb(255, 0, 0)">text</color> here</t>
```

It is difficult to change it later on so we should agree on it before releasing it.

zachberry:
@vutoan1245 I agree with your markup although is any valid color syntax supported (like `hsl` and hex?). If yes, that's great. If not, I think we should just use hex values (like `#FF0099`).

vutoan1245:
@zachberry It supports all color formats that are accepted by html inline style (`style="color:blue"`). I will switch to hex.
Also, I realized there is an html input for color (`<input type="color" />`) Should I use this instead? The color picker is different between browsers and operating systems.

zachberry:
@vutoan1245 We could use that for the custom color perhaps? But since you can't really customize it I think we should stick with your color picker for now. Also, supporting all color formats is good, so you can leave it like that.

vutoan1245:
@zachberry ok, I will leave it like that.

vutoan1245:
I added a color input instead of a text input for custom color. The browser color picker is powerful but difficult to use. Tell me your feedbacks/preferences.

For Chrome OSX:
Screen Shot 2020-08-18 at 1 09 20 PM

zachberry:
@vutoan1245 I think we still need the hex input. Something like `( ) [ hex input ] [ OK ]` - The `( )` would be a circular button with a background set to the current value of the hex input with an eye dropper icon inside of it (I can provide if you want). Clicking on that eye dropper circle button would then open the browser native color picker. Is that possible or is it too complicated to wrangle with the different browsers? I haven't looked outside of Chrome.

zachberry:
Closing in favor of https://github.com/ucfopen/Obojobo/pull/1545

PR #1460, opened 2020-08-11
Added option for tables to have cells with flexible width

### LINKED ISSUE #1400: Option to not have fixed-width table
Based from an instructor request. It would be helpful to have an option to disable the table from having fixed-width columns. This can be useful depending on the type of content input into the table.

maufcost:
Added the option for tables to have cells without fixed width.
Fixes #1400


maufcost:
That's great! I will add the margin and get back to you soon!

jpeterson976 (approved):
LGTM! I agree that a horizontal scroll could be a good addition, though perhaps for a future issue.

Issue #722, opened 2019-02-19
Refactor Table UI in the editor


qwertynerd97:
Right now the UI for the table is very basic and should probably be cleaned up to match the design of Obojobo.  We should add a delete button for the whole table, and refactor the placement of the row and column delete buttons, as well as the buttons below the table

zachberry:
Some mockups:

First, the button with the menu closed. Left is current, right is the mockup:

![image](https://user-images.githubusercontent.com/73479/90258128-1a4b2300-de16-11ea-954a-e5780f72ff6f.png)

Next, the menu open - Left is current, right is the mockup (showing the first item being hovered over):

![image](https://user-images.githubusercontent.com/73479/90258196-2df68980-de16-11ea-9ada-c4de5a39d7c8.png)

zachberry:
I think this menu would also benefit with some icons, but we can spawn that into a new issue

PR #1462, opened 2020-08-17
Updated input styles that were altered with Chrome 83

### LINKED ISSUE #1352: Update input styles to work with Chrome 83 updates
Chrome 83 changed the default design of several form elements. I noticed a few places where this is causing issues for us:

Table button no longer at the correct size:

Screen Shot 2020-05-29 at 10 08 25 AM

Table button menu reverted to gray, should return to white (Also, not sure if the font was reverted or not, but the font should be set to `$font-default`):

Screen Shot 2020-05-29 at 10 08 33 AM

Inputs (at least the ones here) now have a 3D border, should be reverted to a flat border:

Screen Shot 2020-05-29 at 10 08 46 AM

These buttons also have reverted to 3D, and need to be returned to borderless:

Screen Shot 2020-05-29 at 10 09 03 AM

The default ::focus has changed from a fuzzy border to a flat thin border. I like this a lot better, but there are cases like this where the focus border is hard to distinguish since the color is close to the button color. I'm not sure of the best solution, may need some playing around to get something workable in all situations:

Screen Shot 2020-05-29 at 10 09 38 AM

maufcost:
Changed the table button back to white.
Removed inputs that had their borders in 3D back to flat borders.
The table button will also be addressed on the issue #722 .
The Show Assessment Overview button got reverted back to its original form (the incorrect red border doesn't appear anymore).
Fixes #1352


maufcost:
Looks good to me!

PR #1468, opened 2020-08-18
[fix] undo/redo are called twice

### LINKED ISSUE #1464: Undo/redo are being called twice
Only happen when using the keyboard shortcut.
Probably relating to `onKeyDownGlobal`

vutoan1245:
Fixes #1464 

Both visual editor and json/xml editor have their own undo/redo shortcut so we don't have to do it ourselves.


iturgeon:
@zachberry please note my review message above

zachberry:
@iturgeon I was able to replicate your issue if I opened up the (...) menu and then closed it. I had to click back in the editor to get the undo to trigger. I think though it's livable for now.

PR #1479, opened 2020-08-20
Fix splitting figures

### LINKED ISSUE #1275: Splitting Figures in the caption results in duplicate figures
Open to debate on this, but imo splitting the content of a figcaption in a figure should result in a Figure and a Text chunk with the remaining text.

Here's an example of inserting a Break inside the caption of a Figure. Currently: 

1. Figure with caption `>>>`
2. Horizontal Line
3. Figure with caption `<<<`

My expected result would be:

1. Figure with caption `>>>`
2. Horizontal Line
3. Text with `<<<`

ctcuff:
Inserting an item when the cursor is inside a figure caption now inserts that item at the end so it'll be moved into a new line.

Fixes #1275.


ctcuff:
New approach: check if the current selection is inside a figure before a node is inserted. If inserting a node causes the figure to be split, we get the figure 2 lines below the current figure and convert it to text.

I want to also note that this issue could possibly be fixed by implementing #1540.

jpeterson976 (dismissed):
LGTM! Tested inserting each node type mid-caption and they all work the way Zach described.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/file-toolbar-viewer.js
```diff
@@ -46,14 +58,38 @@ const FileToolbarViewer = props => {
...
+						editor.children.length !== prevChildrenCount + 1
+					) {
+						Transforms.setNodes(
+							editor,
+							{
+								type: TEXT_NODE,
+								content: {}
+							},
+							{
+								at: [path[0] + 2]
```
zachberry, line 84: This breaks when splitting a figure inside a QuestionBank Question (for example). I think that's because while the path might be `[4]` on a page, inside a QB Question it might be `[4, 1, 2]`. In that example you'd need to transform the node at `[4, 1, 4]`

zachberry: This code is also very specific, need to make sure we make an issue to modularize this

zachberry (changes requested):

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/file-toolbar-viewer.js
```diff
@@ -46,14 +58,44 @@ const FileToolbarViewer = props => {
...
+				name: item.name,
+				action: () => {
+					const path = Editor.path(editor, editor.selection, { edge: 'start' })
+					const prevChildrenCount = editor.children.length
+
+					Transforms.insertNodes(editor, item.cloneBlankNode())
+
+					// Since inserting a node inside a figure caption can sometimes
+					// cause figure to be duplicated after the caption is split, we
+					// need to convert the duplicated figure below to a text node.
```
iturgeon, line 72: @ctcuff 💯 💯 💯

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/toolbars/file-toolbar-viewer.js
```diff
@@ -46,14 +58,44 @@ const FileToolbarViewer = props => {
...
+					if (
+						containsFigureNode(selectedNodes) &&
```
iturgeon, line 74: Unfortunately this is counterproductive to our efforts to isolate the chunks/modules from one another.

I'd prefer we ask the selected nodes if they need this behavior OR just let the node handle this behavior:

For instance, perhaps each node can implement handleInsert - check for nodes that do, and execute it. pass the path and editor and selected nodes.  Then the node itself could handle the special requirements

This will be needed for iFrame, Materia, and Figure, so it makes sense to not just look for Figure nodes.

iturgeon (changes requested):
Tested in the browser, works fine. 

Needs changes to be more future proof and handle IFrames & Materia Captions

iturgeon:
Updated with a modularized implementation

zachberry (changes requested):
I really like this, I'm happy to move forward with some more modularization!

The only thing I experienced that felt unexpected is if you insert at the beginning of the caption of a figure node. My expecation is that it would become

```
[Figure]
[Inserted Item]
[Figure caption converted into regular text]
```

However the inserted item was placed above the figure:

```
[Inserted item]
[Figure with caption]
```

I'm marking as requesting changes in case we want to handle that now, but I'm fine moving into a new issue as it's not critical (if so I'll appprove).

zachberry (approved):
Marking as approved - my comments from my last review have been made into a new issue - https://github.com/ucfopen/Obojobo/issues/1765

PR #1486, opened 2020-08-25
Link popup menu alignment with hanging indent

### LINKED ISSUE #1477: Hanging indent causes link popup menu to display incorrectly
If the href link popup is displayed in a text chunk that's using hanging indent - the background doesn't render properly.

This is a result of `body *[data-hanging-indent='true']` applying to the popup when it shouldn't. 

I believe this is the style responsible.  Consider changing it so it only applies to the text (so we don't have to do this again) or adding a style to the popup menu to override?

https://github.com/ucfopen/Obojobo/blob/dev/12-ammolite/packages/app/obojobo-document-engine/src/scss/_indents.scss#L2-L5

jpeterson976:
Resolves #1477 

Properly aligns the link text in the href popup menu. The hover background is still offset to the left, though it's just barely noticeable.
Hover background offset

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/marks/link.scss
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/marks/link.scss
@@ -85,3 +85,11 @@
 		}
 	}
 }
+
+*[data-hanging-indent='true'] .editor--components--mark--link .href-menu {
+	overflow: hidden;
+
+	a {
+		text-indent: 2rem;
+	}
+}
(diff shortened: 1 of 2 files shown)


ctcuff (dismissed):
LGTM!

zachberry (approved):
I committed a change - I found a way to keep the border intact in both hanging and not-hanging paragraphs

zachberry (approved):

vutoan1245 (approved):
Looks good.

iturgeon (commented):
Looks Good here

PR #1490, opened 2020-08-27
Fix heading menu in Safari

### LINKED ISSUE #1471: The Heading menu in the editor does nothing in Safari
You can click on the items in the menu but it doesn't change the text

ctcuff:
The problem was that `onClick` fired after `onBlur`. This meant that the menu would close without doing anything. Changing the menu buttons `onClick` to `onMouseDown` lets the event fire before `onBlur`. However, this means that we need to manually listen for an `Enter` keypress since pressing Enter called `onClick`.

Fixes #1471.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/toolbars/paragraph-styles.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/toolbars/paragraph-styles.test.js
@@ -154,7 +154,7 @@ describe('Paragraph Styles', () => {
 				preventDefault: jest.fn()
 			})
 
-		expect(mockChangeToType).toHaveBeenCalledWith(TEXT_NODE)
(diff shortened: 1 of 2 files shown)


zachberry:
Forgot to say - I tested in Chrome, FF and Safari. I only confirmed the focus issue with Chrome, not sure how FF and Safari handle it.

ctcuff:
Quick note slightly related to @zachberry's 2nd point: If you use the arrow keys to navigate through the menu, then close the menu, focus remains on the last item you navigated to when you re-open the menu. This occurs on dev/13 but I'm not sure if this is something worth looking in to.

zachberry:
@ctcuff I'm thinking we spawn into a different issue as it's relatively minor, and your fix is working

zachberry (approved):
Tested in FF, Chrome, Safari, works!

jpeterson976 (approved):
Also tested on Chrome, FF, and Safari, LGTM!

PR #1491, opened 2020-08-28
Removes "undefined undefined" from /drafts/:draftId/revisions/:revisionId

### LINKED ISSUE #1478: /drafts/:draftId/revisions/:revisionId returns "undefined undefined" for "userFullName", missing "userId"
I noticed this when asking for /drafts/:draftId/revisions/:revisionId -

image

Also, I see that the objects returned in /drafts/:draftId/revisions include "userId". I figure we should add it to the .../:revisionId endpoint for consistency

ctcuff:
Also adds `userId` to `/drafts/:draftId/revisions/:revisionId`.

Fixes #1478.


vutoan1245 (dismissed):
Looks good.

zachberry (changes requested):
Now I'm not getting `userFullName` in my query result. This is for user "1" and I can confirm they have first_name and last_name in the users table.

iturgeon (approved):
looks good.  Seems like something here got merged out in the past when drafts_content got user_ids.

PR #1494, opened 2020-08-31
Shows assessment scores and attempts in nav

### LINKED ISSUE #541: Show assessment scores and attempt numbers in nav
_Friday Jun 15, 2018 at 14:29 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Document-Engine/issues/245_

----

See this older mockup (also in #329 ):

32524077-f2e09ab2-c3d2-11e7-8a80-3ee5bfa98075

Would like

1. Assessment scores to display in the nav so that a student doesn't need to navigate to the assessment section to see their score
2. Attempt counts to show in the nav (not shown in mockup). #540 may be the solution to this.

jpeterson976:
Resolves #541 

Adds score and attempt information for assessments into the nav menu.


packages/app/obojobo-document-engine/src/scripts/viewer/components/nav.js
```diff
@@ -143,12 +145,22 @@ export default class Nav extends React.Component {
...
+				</li>
+			)
+		} else {
+			return (
+				<li key={index} onClick={this.onClick.bind(this, item)} className={className}>
+					{this.renderLinkButton(item.label, ariaLabel, isItemDisabled, item.id)}
+					{lockEl}
+				</li>
+			)
+		}
```
ctcuff, line 163: _SUPER_ minor nitpick but you could simplify this a bit:

```javascript
return (
	<li key={index} onClick={this.onClick.bind(this, item)} className={className}>
		{this.renderLinkButton(item.label, ariaLabel, isItemDisabled, item.id)}
		{item.flags.assessment ? (
			<span className="assessment-info">{this.getAssessmentInfo(item)}</span>
		) : null}
		{lockEl}
	</li>
)
```

vutoan1245 (approved):
Looks nice. I don't see any problem.

zachberry (changes requested):
This hardcodes some assessment logic inside the nav, which I would rather move out. Ideally I think the assessment's viewer.js's `getNavItem` could return a line of text instead of generating the attempts line inside nav. However, that would be a pretty significant refactor of the nav to get that to work, and there's lots of existing changes I want to make in the nav so it would be a can of worms.

So all that said, I think this is a nice improvement - I'm thinking we push ahead and I'll make an additional issue for the refactor for us to tackle later.

I did find a few things:
1. It's possible to have unlimited attempts - you can make that happen by editing a module in XML, finding the Assessment tag, and changes attempts to `unlimited`. Then the number of attempts in the viewer is set to Infinity. So you'll want to check for that case and write "Unlimited" instead.
2. My preference would be to keep the score line the same even when the Assessment section is clicked on and turns the label to bold blue. Mockup:

image

iturgeon:
someone's using windows :) @jpeterson976 

`.DS_Store` file made it in to a commit.  remove and add `.DS_Store` to the .gitignore

iturgeon (approved):

PR #1495, opened 2020-09-01
Youtube clicking issue

### LINKED ISSUE #1473: Can't bring up the More Details menu (...) for YouTube chunks
Clicking on the YouTube chunk never seems to show you the More Details menu

vutoan1245:
Fixes #1473 
Fixes the same issue when clicking on the non-text part of Question and QuestionBank


zachberry (changes requested):
I found a few issues:

There's a focus issue when clicking on parts of a Question, shown here:



Second - clicking on a YT video to get the `(...)` menu to show up only works when there's no video. Once a video is set the video starts playing when you click but the menu no longer shows up. Ideally we could still play the video but also somehow register the click to get the `(...)` menu to display.

vutoan1245:
@zachberry 
About the Youtube issue you mentioned, apparently the iframe blocks the 'click' event. I cannot by any solution except disabling the iframe. The (...) menu will show up but the video will not play. What do you think?

zachberry:
@vutoan1245 In that case, perhaps we should use the thumbnail image instead of actually embedding the iframe. Later we can add a button to play the video. You can get the thumbnail of a video with `//img.youtube.com/vi/<VIDEO-ID-HERE>/maxresdefault.jpg` - for example `//img.youtube.com/vi/dQw4w9WgXcQ/maxresdefault.jpg`

iturgeon:
Needs to be re-evaluated due to age, marking as a draft

FrenjaminBanklin:
The specific issue of the detail menu button not appearing after adding a video URL does seem to have been resolved at some point, so I think I may close this. There are a few other things I noticed which I will make a new issue for.

PR #1497, opened 2020-09-03
Unit tests for Module Repository

FrenjaminBanklin:
Added unit tests for everything relevant in the repository - copied from the private collections branch and rearranged/removed tests as needed to cover current code, minus all the other new additions from that branch. Modified existing code as needed to make it more testable, standardized filenames to '.jsx' for files with JSX and '.js' for files without.

Since this was copied from another branch with additional changes that were also fully tested, there may be some leftover code for the additional features that I missed when I was cleaning up.


zachberry (dismissed):
Nice work!

packages/app/obojobo-repository/server/models/draft_summary.test.js
```diff
@@ -0,0 +1,516 @@
...
+		return DraftSummary.fetchById('mockDraftId').catch(err => {
+			expect(logger.error).toHaveBeenCalledWith('fetchById Error', 'not found in db')
+			expect(err).toBe('Error Loading DraftSummary by id')
+		})
+	})
+
+	test('fetchByUserId generates the correct query and returns a DraftSummary object', () => {
+		db.any = jest.fn()
+		db.any.mockResolvedValueOnce(mockRawDraftSummary)
+
```
iturgeon, line 123:
```suggestion
		expect.hasAssertions()
```

packages/app/obojobo-repository/server/models/draft_summary.test.js
```diff
@@ -0,0 +1,516 @@
...
+				ON drafts_content.user_id = users.id
+			WHERE
+				drafts_content.draft_id = $[draftId]
+			ORDER BY
+				drafts_content.created_at DESC
+			LIMIT $[count];
+		`
+
+		return DraftSummary.fetchAllDraftRevisions('mockDraftId').then(history => {
+			expect(db.any).toHaveBeenCalledWith(query, {
```
iturgeon, line 200: Quick idea, not sure, this maybe could use an 'expect string contains' against the important bits of the query?

packages/app/obojobo-repository/server/models/drafts_metadata.test.js
```diff
@@ -0,0 +1,79 @@
...
+	test('saveOrCreate returns a DraftsMetadata object', () => {
+		db.none = jest.fn()
+		db.none.mockResolvedValueOnce(mockRawDraftsMetadata)
+
+		const draftMetadata = new DraftsMetadata(mockRawDraftsMetadata)
+
+		expect.hasAssertions()
+
+		// Trying to match whitespace with the query that's actually running
+		const query = `
```
iturgeon, line 46: another idea for these (if string contains isn't any good), use a snapshot test.

packages/app/obojobo-repository/server/services/search.test.js
```diff
@@ -0,0 +1,71 @@
...
+		const mockResponse = [{ ...mockRawUser }, { ...mockRawUser, id: 2 }, { ...mockRawUser, id: 3 }]
+
+		db.manyOrNone.mockResolvedValueOnce(mockResponse)
+
+		db.taskIf = jest.fn()
+		db.taskIf.mockImplementation(cb => cb(db))
+
+		expect.hasAssertions()
+
+		const manyOrNoneQueryString = `SELECT
```
iturgeon, line 38: same notes on testing a match of this query string as above

packages/app/obojobo-repository/setupTests.js
```diff
@@ -0,0 +1 @@
+require('obojobo-lib-utils/test-setup-chunks') // setup enzyme
```
iturgeon, line 1: just making sure this is needed (if so maybe we should rename test-setup-chunks)

packages/app/obojobo-repository/shared/actions/dashboard-actions.js
```diff
@@ -229,7 +229,7 @@ const promptUserForModuleFileUpload = async () => {
 
 const moduleUploadFileSelected = (boundResolve, boundReject, event) => {
 	const file = event.target.files[0]
-	if (!file) boundResolve()
+	if (!file) return boundResolve()
```
iturgeon, line 232: whoopsie.  Yay tests!

iturgeon (approved):
Looks pretty awesome to me

packages/app/obojobo-repository/server/models/collection.test.js
```diff
@@ -0,0 +1,150 @@
...
+			id: 'mockCollectionId',
+			title: 'mockCollectionTitle',
+			user_id: userId,
+			created_at: new Date().toISOString()
+		}
+
+		db.one.mockResolvedValueOnce(mockNewRawCollection)
+
+		return CollectionModel.create('', userId).then(model => {
+			expect(model).toBeInstanceOf(CollectionModel)
```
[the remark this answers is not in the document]
FrenjaminBanklin: Ah, leftovers - when I made changes to the `RepositoryCollection` model in the private collections branch I also added logging for creation collection - that part got removed, I just forgot to rename the test. Currently, there's no logging on creation of a collection.

Writing a separate test to make sure `db.one()` is being called correctly is a fair point, though I think the `expect`s here match the intention of making sure that calling `.create` with no title will still return a collection model with the correct defaults.

packages/app/obojobo-repository/server/models/draft_summary.test.js
```diff
@@ -0,0 +1,516 @@
...
+		WHERE drafts.deleted = FALSE
+		AND ${whereSQL}
+		WINDOW wnd AS (
+			PARTITION BY drafts_content.draft_id ORDER BY drafts_content.created_at
+			ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
+		)
+		ORDER BY updated_at DESC
+	`
+
+	const checkAgainstMockRawSummary = summary => {
```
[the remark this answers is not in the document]
FrenjaminBanklin: Good point. Easy change.

packages/app/obojobo-repository/server/models/draft_summary.test.js
FrenjaminBanklin: Oops. Added.

packages/app/obojobo-repository/server/models/draft_summary.test.js
FrenjaminBanklin: I think I might have tried that, but it ended up feeling a bit kludgy for queries where there was a lot to look for. I had kind of hoped there was some jest magic to ignore initial indentation, but that didn't pan out either.

packages/app/obojobo-repository/server/models/drafts_metadata.test.js
FrenjaminBanklin: I'm mostly writing out queries like this to pass them into a `toHaveBeenCalledWith` to make sure we're sending the right queries to our `db` calls - maybe it's a feature of snapshots I'm not familiar with, but I'm not sure if they'd be applicable here.

packages/app/obojobo-repository/setupTests.js
FrenjaminBanklin: Not sure about the name, but some of the tests use Enzyme so this'll definitely be necessary.

FrenjaminBanklin:
There's still a weird Async timeout error coming from the `repository library route › get / returns the expected response` test that I haven't been able to figure out. If run in isolation (or even just alongside the other repository tests) it works fine, but when all of the Obojobo tests run it will sporadically time out - usually only just the first time the tests are run, and then a subsequent run will succeed without a hitch.

Another weird thing - something about this test failing bleeds into the next (completely unrelated) test, causing failures in that as well - although Jest will report the failure as coming from the first test. It's all very strange.

There's got to be some way of preventing this altogether, but since I can't even figure out how to reproduce it reliably I'm not sure how to approach it.

packages/app/obojobo-repository/server/models/drafts_metadata.test.js
iturgeon: Don't forget you can just extract a single argument to a function `db.one.calls[0][0]` (first argument of first call).

Another option is `.toHaveBeenCalledWith(expect. stringContaining('query'), {})`

perhaps `expect(db.one.calls[0][0].replace(/[\W\t\n\r]+/g, ' ')).toBe(expectedQuery.replace(/[\W\t\n\r]+/g, ' '))`

FrenjaminBanklin: That could work. Might want to put it behind reusable function to denastify it for multiple uses, though. Probably a bit late to retrofit the tests now, but perhaps for a refactor later and/or all future tests doing string comparisons like this.

iturgeon (approved):

PR #1508, opened 2020-09-10
Changed the default option for the fixed-width table cells

### LINKED ISSUE #1400: Option to not have fixed-width table

maufcost:
Changed the default option for the fixed-width table cells and removed unnecessary comments.
Fixes #1400


jpeterson976 (approved):
LGTM!

PR #1519, opened 2020-09-12
Fix long text breaking auto width tables

### LINKED ISSUE #1518: dev 13, auto width table cells highly likely to cause side scrolling
The auto width cells feature implemented in Dev/13 are capable of expanding the width of the viewer.

It breaks some of the design of the viewer and most certainly won't be friendly to small screens.

zachberry:
Fixes #1518 

Adds the break word rule to the auto-width table, fixing an issue with really long strings of text extending the table past the available space


iturgeon (approved):
Looks like a on ok compromise. The browser just isn't super good at predicting the best course of action. The cells with the least text will break before we'd like, but at least the rest of the page isn't super wide.

iturgeon (approved):
Looks pretty good in my tests.

PR #1520, opened 2020-09-14
adds onStartAttempt trigger to lock nav

### LINKED ISSUE #1503: Lock Navigation During Attempts missing 'onStartAttempt' trigger
The 'Lock Navigation During Attempts' feature automatically adds some triggers to ensuring that when an attempt is started the navigation is locked. However it's missing an 'onStartAttempt' trigger which should run the lock navigation action. If you don't apply the 'lock navigation' action on your Start Assessment button then students can leave the assessment.

jpeterson976:
Resolves #1503 

Added a trigger to lock the nav when an attempt is started.

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/sub-menu.js
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/sub-menu.js
@@ -162,11 +162,9 @@ class SubMenu extends React.Component {
 	}
 
 	lockValue(content) {
-		const startAttemptLock = hasTriggerTypeWithActionType(
-			content.triggers,
-			'onNavEnter',
-			'nav:lock'
-		)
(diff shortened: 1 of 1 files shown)


zachberry (changes requested):
So this is unfair, because this is perfect and not related to your code. But in re-reading the label of the toggle switch it says "Lock navigation during Attempts" but with everything that it's doing it's really "Lock navigation when visiting the assessment page and during attempts" - That's useful when creating "Chen" style pre-test modules, but here I'm thinking we should simplify and only handle locking/unlocking of nav when the attempt is started.

So, all that said - I think we should remove the nav:enter and nav:exit actions that this toggle adds - it now should only add the assessment start and assessment end events.

zachberry (approved):
Awesome, thanks! LGTM.

ctcuff (approved):
Looks good!

PR #1523, opened 2020-09-14
Updates the table padding added for when scrollbars are used to recalculate on page resize

### LINKED ISSUE #1522: Add/remove additional scrollbar padding scrollbars for auto-width tables on page resize
The additions in #1519 to account for scrollbars happened on component mount, but if the screen is resized the need for extra padding is not recalculated.

zachberry:
Fixes #1522


iturgeon (approved):
Solves the intended issue for me.  Maybe a width change aware higher order component might be useful in the future.

PR #1524, opened 2020-09-14
Made the left sidebar sticky and scrollbar appear after top navbar

### LINKED ISSUE #1263: Force nav bar in visual editor to stay at the top of the editing area

maufcost:
The left sidebar now accompanies scrolling and the page's scrollbar appears after the top navbar (and not under).
Fixes #1263


zachberry (changes requested):
Seems great but there's one case I forgot about that is unfortunately going to be annoying. The nav should scroll in cases where your window isn't very tall and you have a lot of pages (you can get this quickly be editing a new Tutorial document). Right now if your page is off the screen you have to scroll down to get past the 'sticky' rule.

The problem however is that by giving the .sticky div some scrolling rules the MoreInfoBox no longer sticks out of the right side of the menu - it's now trapped inside the nav box.

May require some creative div hackery. Perhaps the nav needs to be in a container wide enough to show both the nav and the MoreInfoBoxes. But that would need to not interact with your ability to edit the left-side of the content, so maybe the nav container has pointer-events: none on it, and the nav within restores that with pointer-events: auto.

maufcost:
I'm working on it. I believe expanding the div won't work because the y-axis scrollbar will also be pushed to the right (as you can see on the image). What I'm trying to do is moving some components around so that `MoreInfoBox` can get out of the sticky rule and be on top of the scrollbar (just like the GitHub home page's left sidebar).

obo
(Ignore the colors. They just help situate myself on a document)

maufcost:
Here are the React module structure of the left sidebar in a nutshell:

Editor-nav > the `li`s you see inside of it > more info box

I found out what was the main issue, and, as far as I know, we have two options:

- Tamper with CSS to make the more info box "_absolutely_" absolute in respect to the entire page. Currently, the `li` that the more info box is present in is the relative so I can't just keep the more info box to being absolute and position it **outside** the li (this option is kind of gruesome because I would need to remove the relative aspect of `li`s and there are other aspects of this page that depends on the li being relative. So, tampering with that brings a lot of other design issues).
- That's the option I am going for: remove the more info box component from inside the `li`s and let it be alongside the editor-nav component, which will make it possible to put the more info box dialogs on top of the scrollbar (in case an overflow-y occurs on the left sidebar - e.g. the tutorial module).

^ This is mainly for bookkeeping, but feel free to let me know if anyone has any thoughts :)

maufcost:
The commit above is what I've been working on for past couple of days, but should _only_ be used for reference in the future (tests and css are still off). I implemented my original approach of changing the lexicographical level of the `MoreInfoBox` so that it could be on top of the sticky left sidebar's scrollbar. This part works fine now without interfering with authors' abilities to access the editor.
However, upon further examination, I found out that even more JSX structural changes should be made to the left sidebar in order for the more info box icons and the box itself to follow the left sticky sidebar.

iturgeon:
Changing ot a draft based on @maufcost's last comments

PR #1539, opened 2020-09-21
Rubric input focus issue

### LINKED ISSUE #1505: Focus is stolen when editing assessment rubric values in the assessment section
Attempting to input, for example, 50% in the first numeric input field in the assessment rubric, doesn't work. After typing "5" the focus is lost and typing "0" does nothing.

I'm sure this is the same problem as other focus-stealing issues we've fixed in the editor for (example) the Question Bank settings.

vutoan1245:
Fixes #1505 

Problem: the onChange function constantly calls `Transforms.setNodes` which causes the input to lose focus.
Solution: save all input values in state and call `Transforms.setNodes` when `editor:save` is triggered. Of course, `editor:save` event needs to be created in `visual-editor.js`.

This is the only solution I can think of. Tell me if anybody has a different idea.


maufcost (approved):
LGTM!

packages/obonode/obojobo-sections-assessment/components/rubric/editor-component.js
```diff
@@ -16,19 +17,36 @@ class Rubric extends React.Component {
 	constructor(props) {
 		super(props)
 
+		const content = this.props.element.content
+		this.state = {
+			passingAttemptScore: content.passingAttemptScore || '',
+			passedResult: content.passedResult || '',
+			failedResult: content.failedResult || '',
+			unableToPassResult: content.unableToPassResult || ''
```
jpeterson976, line 25: Since `failedResult` and `unableToPassResult` are 0 by default, it seems to evaluate them as falsy and chooses `''` instead for those values. This can cause issues if those values are not overridden in the editor and you take an assessment that meets failing criteria. 

I think that instead of `''` we should give them the current default values of 100 for `passingAttemptScore` and `passedResult`, and 0 for `failedResult` and `unableToPassResult` to prevent that error from occurring.

jpeterson976 (changes requested):
Works great! There's just a bit of JS weirdness in setting up the state that can potentially cause an error.

zachberry:
I updated this for dev/16 and I think I addressed @jpeterson976 's comments.

@vutoan1245 's solution to update the state of the rubric UI changes was using a new editor:save event, which I like, however since this PR was opened code to saveDraft/saveModule has been changed and that no longer works. Additionally, I wanted to implement a way for editor:save to wait for any listening function to complete its tasks before allowing the save to complete, and this was going to be non-trivial to implement.

Instead, I'm now saving the state of the rubric when a user mousedowns away from the rubric. This does result in some page jumping around after clicking away from the rubric. I believe it's the lesser of two evils for now. I think we should probably reconsider having this UI in slate and instead move it to a dialog. Originally we didn't do that because the mods dialog would require that we have a dialog in a dialog. But I still think we could get around that, and if not, dialog in a dialog is still better than what we have now imo. But I'm willing to change my mind.

Ready for re-review now

jpeterson976 (approved):
LGTM, the only things that I found are known issues. I think I agree with Zach's thoughts about this, too.

ctcuff (approved):
LGTM! To mention Zach's idea, I think this would be a great use case for a modal!

PR #1545, opened 2020-09-29
Color Text

### LINKED ISSUE #1162: Colored text

vutoan1245:
Resolves #1162 

Problem with git, I have to submit this PR again. Old PR: https://github.com/ucfopen/Obojobo/pull/1459

Screen Shot 2020-09-29 at 10 44 13 AM


zachberry:
I messed things up and added some tweaks to this:

* Color Picker now allows you to input any type of valid CSS color string
* Color Picker custom color icon only updates when a valid color is inputted
* Inputting "" will remove any color content
* The swatch for the color of the selected text will be highlighted, if it exists
* Color Picker opens in expanded mode if an expanded or custom color was chosen
* Hitting enter in the custom color text input submits the color and closes the dialog (does not work in FF)
* Some small CSS tweaks
* Updated the Expand button to use $color-action-bg, also updated the Link Menu to use this as well
* Repositioned the color input control to take up the same location as the color icon, resulting in the color input control to show up near the icon
* Added a keyboard shortcut - Meta+Shift+P
* Moved the color picker toolbar button to the right

An issue I wanted to fix but ran out of time - The menu isn't accessible. We need to make the color-cells tab-able. I'm not sure if we should do that now or do it in the next release.

zachberry:
Updated to allow the color swatches to be tab-able (they've been switched to buttons)

vutoan1245:
The new changes look incredible.
I found and fixed a crashing issue when opening the color picker for non-text.

jpeterson976 (dismissed):
Works great! I really love the color picker UI, I think it looks great. The only thing I noticed is that if you change the color of text in a question, it doesn't appear as that color in the import question menu, but the imported question itself does. I think that this could be made into a separate issue though, so LGTM!

ctcuff (commented):
Looks great so far, this is gonna be a really cool feature to have! I did notice a few small issues:

When you have multiple colors in a heading:
Screen Shot 2020-10-23 at 11 30 21 AM
 pressing enter at the beginning of that heading splits it into multiple lines of text:
Screen Shot 2020-10-23 at 11 30 36 AM
This might be worth fixing in another issue since the same thing happens with bold/italicized text.


Not sure if this is intentional, but colored text also appears in the nav, which may be an accessibility issue:

Screen Shot 2020-10-23 at 11 31 32 AM

It may also be worth adding a color text section to the Visual Editor Tutorial

vutoan1245:
@jpeterson976 @ctcuff 
Nice catch! Should be good now.

ctcuff (changes requested):
The implementation looks great and so far the requested changes have been fixed! After I merged in dev/16, I got a few errors though:

One error occurs when you try to color text that's just spaces. When you switch to the XML viewer, you end up with this:
```xml
<Text id="f72e7df5-766b-4e34-a003-90e0b80e373b">
    <textGroup>
        <t align="left" indent="0">
            <color text="#cc0000">
            </color>
        </t>
    </textGroup>
</Text>
```
When you switch back to the Visual Editor, you get an error message:
Screen Shot 2021-02-01 at 2 47 54 PM 
I think this may be caused by the empty `<color>` tag.

This one is less of a problem, but when you add colored text anywhere in the editor then try to preview the module, the `<color>` tag gets left in the DOM and you get a warning in the console:
```
Warning: The tag <color> is unrecognized in this browser.
If you meant to render a React component, start its name with
an uppercase letter.
```
The HTML looks something like this:
```html
<h1>
    <span>
        <color>
            <span>
              .
              .
              .
            </span>
        </color>
    </span>
</h1>
```

We should also probably disallow coloring HTML. It works just fine in the editor, but if you switch to the XML or JSON viewer, then switch back to the visual editor, the color gets lost.

zachberry:
I created #1694 to address some of the issues that @ctcuff found. I also found that without that PR the generated XML/JSON could become very complex, and it made sense to simplify this before adding more code. So I merged that into this - #1694 will have to be merged in first before this.

This latest change adds color is officially a part of the StyleOrder in StyleableTextRenderer, which resolves other issues

maufcost (dismissed):
After merging in #1694 , and re-testing all the issues found above, I applied colors to headings, questions, solutions, explanations, buttons, codes, table, etc, and they were all correctly updated from the visual editor to the viewer. I also made changes using only the JSON and XML editors, and they all looked good to me. **Amazing work everyone!** LGTM!

Just one issue (that is more aesthetical and a matter of person opinion and does not impact any functionality whatsoever):
I think the current set of colors that appear when we click on the color picker icon on the toolbar is kind of too dark. For example, when you think about the orange color, the first thing that comes to mind is this:

Screen Shot 2021-03-09 at 10 33 03 AM

That's how it is right now:

Screen Shot 2021-03-09 at 10 22 58 AM

zachberry:
> After merging in #1694 , and re-testing all the issues found above, I applied colors to headings, questions, solutions, explanations, buttons, codes, table, etc, and they were all correctly updated from the visual editor to the viewer. I also made changes using only the JSON and XML editors, and they all looked good to me. **Amazing work everyone!** LGTM!
> 
> Just one issue (that is more aesthetical and a matter of person opinion and does not impact any functionality whatsoever):
> I think the current set of colors that appear when we click on the color picker icon on the toolbar is kind of too dark. For example, when you think about the orange color, the first thing that comes to mind is this:
> 
> Screen Shot 2021-03-09 at 10 33 03 AM
> 
> That's how it is right now:
> 
> Screen Shot 2021-03-09 at 10 22 58 AM

Thanks @maufcost !  I agree with you, however the reason we have the dark colors is to maximize color contrast against a white background for accessibility reasons. The default colors all have a strong contrast ratio other than yellow, which is difficult.

I do feel perhaps the red is too dark, it's easier to see it as red when you pick the one swatch down. I did a test - top row is the first set of colors, the second row is the second set of colors:

image

My thought is we push ahead with this as is, but I'm open to spending some time to tweaking the colors. I'd still want to maintain the strong color contrast, but we can shift the shades around.

jpeterson976:
Tested this a bit, and seems like pretty much all the requested changes are addressed - though on my end I still get the `format unexpected` error when swapping from XML to the visual editor if there's a string of just spaces with color, which prevents saving while using the XML editor. I'm not sure how simple this would be to fix, though it can maybe be turned into its own issue since I imagine it's probably not going to happen in the wild super often.

zachberry:
@jpeterson976 Agree - happens as well with `<b> </b>`, so I'm going to file as a new bug

PR #1558, opened 2020-10-01
Fixes the assessmentScoreId in the assessment:attemptScored event

### LINKED ISSUE #1556: Update assessment:attemptScored event to fix assessmentScoreId
In this event `assessmentScoreId` is incorrectly returning the attempt data object instead of the Id. This information is useful but since we're close to release I'm thinking we revert the value back to the expected value and make a new issue to add in this information.

zachberry:
Resolves #1556


iturgeon (approved):

jpeterson976 (approved):
LGTM!

PR #1566, opened 2020-10-02
adds github action for publishing npm releases, adds yarn cache to gh actions

iturgeon:
Adds action that'll publish to NPM when a release is published

Updates the main test action to use cache on the yarn packages directory to attempt to speed up package downloading on the action runners.


.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,25 @@
+name: NPM publish release
```
zachberry, line 1: This [example](https://github.com/actions/starter-workflows/blob/main/ci/npm-publish.yml) creates a build job which I like - what do you think? I'm okay moving ahead as is however.

zachberry (changes requested):
I had a question posted in the comments. Otherwise, this is a bit outside my area of expertise but it seems like it would do the job as expected.

.github/workflows/npmpublish.yml
iturgeon: I'm not opposed, but don't think it's necessarily _better_ in the case where we only publish to npm

zachberry (dismissed):

maufcost (dismissed):
This looks pretty good to me as well. I've compared it with other workflows on GitHub, and it  seems pretty solid yaml- and command-wise.

iturgeon:
I'm inclined to connect this to the tests.yml suite...  That runs every pr and when it runs on a tag - it publishes to the docker registry.  It makes sense to me that we publish to npm in that same flow?

By changing to that flow, it'd publish all tags to npm (including alpha, rc, etc) instead of just when we publish a release on github.

PR #1567, opened 2020-10-02
adds prettier github action

iturgeon:


.github/workflows/prettier.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull_request:
+  # for pull requests from forks
+  pull_request_target:
+  # for any pushes that may sidestep pr workflow
+  push:
+    branches:
+    - main
+    - master
+    - dev/*
+    - hotfix/*
```
zachberry, line 13: I've seen people use `feature/*` before

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.head_ref }}
+
+    - name: NPM - Test, Build, Deploy
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.11.1
```
zachberry, line 43: If we update our node version will we remember to update this as well? Pretty sure the answer is 'no', but is it possible to grab this from the package json?

zachberry (changes requested):
I tested pushing a commit with some un-prettier files and it seemed to work great. Not sure if I'm missing some edge cases, but it seems like it's working. I did have some questions on the config file, so marking as 'request changes'

.github/workflows/prettier.yml
iturgeon: sounds good, I'll def add that.

.github/workflows/prettier.yml
iturgeon: I've been pretty good about orchestrating node version changes in the past (usually a global search of the current version). But you're right, a likely situation that we miss this file.  

I can't think of any sustainable way to script changes to node versions that wouldn't just move the problem somewhere else.

PR #1568, opened 2020-10-02
adds manual action for tagging a release

iturgeon:


.github/workflows/tag.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.11.1
+        registry-url: https://registry.npmjs.org/
+
+    - name: Cache Yarn Deps
+      uses: c-hive/gha-yarn-cache@v1
+
+    - run: yarn install --frozen-lockfile
+    - run: yarn lerna version --no-push --no-commit-hooks --force-publish --no-git-tag-version --yes ${{ github.event.inputs.newTag }}
```
zachberry, line 29: This is missing the `--sign-git-commit --sign-git-tag` options from `release:tag`, just making sure that's expected. Also, with this action added, should we remove the `release:tag` command? Or any other publish-related commands?

zachberry (changes requested):

.github/workflows/tag.yml
iturgeon: We'd have to provide a secret GPG key to sign (actions can't yet sign using the keys provided by users to sign commits via the github GUI)  RE: https://github.com/actions/runner/issues/667

For the sake of progress, I think I'm willing to give up signed tags 😢

PR #1572, opened 2020-10-03
github test action optimization

iturgeon:
* uses cache for yarn deps (looks like a ~1 minute run time reduction)
* adds `--frozen-lockfile` to yarn install
* removes `npm install -g yarn` as yarn comes with actions/setup-node@v1


zachberry (approved):
This LGTM, although I haven't tested this action. However since it deals with tests, if there is a problem it should only impact dev branches and should be easy to rectify.

zachberry (dismissed):
I haven't tested but in looking into gha-yarn-cache and --frozen-lockfile those both seem like good options to enable

maufcost (approved):
Everything worked normally on my end. Good job, Ian!

iturgeon:
This needed to be updated to merge into dev/17, but it looks like there's some yarn install errors

```
warning workspace-aggregator-3d020564-ccc8-4d58-bc94-9e921855b6a1 > obojobo-express > webpack > watchpack > chokidar > fsevents@2.1.3: "Please update to latest v2.3 or v2.2"
```

iturgeon:
Fixed the issue by updating a few of the pinned dependencies in the yarn.lock file

PR #1583, opened 2020-10-05
Cleaned up intersection observer when the component unmounts

### LINKED ISSUE #1552: Clean up ViewerApp IntersectionObserver on umount
* Need to remove the IntersectionObserver on viewer app's unmount method
* Need to add comments to `startObserving...` since it seems like this would create an IntersectionObserver every time but in reality only does if `visualFocusTarget` is set.

maufcost:
Cleaned up the intersection observer at the viewer app when it unmounts and added the requested comments to avoid future confusion.
Fixes #1552


zachberry (dismissed):
Made a slight tweak to one of the comments, and then noticed the `threshold` value from a previous commit was misspelled so I fixed it here. The default `threshold` value is 0, so the spelling mistake didn't impact anything.

LGTM!

ctcuff (dismissed):
LGTM!

PR #1585, opened 2020-10-07
Triple-clicking selection issue

### LINKED ISSUE #1550: Triple clicking to select the entirety of a line and then typing jumps the selection to the next node
Best shown here:

We've had issues with this before with Slate - it could still be a slate issue or perhaps something to do with breakToText?

vutoan1245:
Fixes #1550 

Note: This is an open Slate issue. I am fine with disregarding this PR and wait for Slate to fix it.

Problem: triple-clicking causes selection to bleed into the beginning of the node below it. It doesn't happen in Firefox.
Solution: if triple-click happens and selection ends at the beginning -> reduce selection by 1
https://docs.slatejs.org/api/transforms#transforms-move-editor-editor-options

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/visual-editor.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/visual-editor.test.js
@@ -10,8 +10,9 @@ import ModalStore from '../../../src/scripts/common/stores/modal-store'
 import Dispatcher from '../../../src/scripts/common/flux/dispatcher'
 import OboModel from 'src/scripts/common/models/obo-model'
 
-import { Editor } from 'slate'
+import { Editor, Transforms } from 'slate'
 import { ReactEditor } from 'slate-react'
+// jest.mock('slate')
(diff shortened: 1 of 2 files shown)


zachberry (changes requested):
This works great except for one case I found - when triple clicking on the last line of a question the selection disappears

vutoan1245:
@zachberry It seems the problem with triple clicking in a question is a separate issue.

zachberry (approved):
I spent a long time seeing if I could come up with another way to handle the bug when selecting the last line of a question but as far as I could tell the selection is simply not set (for whatever internal Slate reason) when triple clicking the last line of a Question. Since this PR otherwise is an improvement I'm thinking we go ahead and ship, and hope that a future Slate release natively fixes the issue and we can remove this workaround.

ctcuff (approved):
LGTM, although there's one super tiny edge case: you can still get multiple selections by triple clicking an empty node:

Screen Shot 2021-02-09 at 2 10 44 PM

Although, approving since I'm not sure if this is something we can fix, may just have to wait for slate to resolve it.

PR #1589, opened 2020-10-12
adds lock env vars to herkou app.json

iturgeon:


zachberry (dismissed):
Was able to build a herkou box and tested completing an assessment

Issue #1483, opened 2020-08-21
Splitting a Figure caption produces multiple lines of text when that text is formatted


zachberry:
Best shown in an example:

![2020-08-21 14 12 39](https://user-images.githubusercontent.com/73479/90924738-00818100-e3be-11ea-99fd-d2c57124ec8f.gif)

My expectation is that the caption text that got converted into regular text would remain one line. I suspect there is code in Text chunk's normalization methods that is doing this.

maufcost:
Just found out another issue that I believe can be within this thread. When you type "return" when your text cursor is right behind the styled text, the figure disappears. 
![obo](https://user-images.githubusercontent.com/39862359/93892259-577abe80-fcba-11ea-9180-95da671ce9c2.gif)

PR #1590, opened 2020-10-12
Bugs in breakToText

### LINKED ISSUE #1482: KeyDownUtil.breakToText doesn't produce expected behavior when handling a non-collapsed selection
You can test this by creating a few Headings, then select a portion of those and hitting enter.

```
Hea[ding text

Heading text

Heading te]xt
```

(`[...]` represents the text selection)

My expectation is when hitting enter the portion between `[...]` will be deleted, the top and bottom heading merged, then the remaining portion after the selection split off and converted into a regular text node:

```
Hea

|tx (<-- This would be regular text)
```

(`|` represents the cursor)

Perhaps we need to change the behavior of breakToText such that it allows Slate to function normally, then it simply converts the last node to a Text chunk type node.

vutoan1245:
Fixes #1482, #1483

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
@@ -1,10 +1,13 @@
-import { Transforms } from 'slate'
+import { Transforms, Point } from 'slate'
 import { ReactEditor } from 'slate-react'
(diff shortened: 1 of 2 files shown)


jpeterson976 (approved):
LGTM! Works great for both headings and figures with my testing.

zachberry (approved):
This is great! LGTM

PR #1592, opened 2020-10-14
Standardized trigger messages in action buttons with undefined ids

### LINKED ISSUE #1511: Custom message for ActionButton if no target id is set
If you edit the onClick trigger of an ActionButton and then create a nav:goto action, but don't supply an ID you get this message:

image

However, if you make a start attempt action, you get this message:

image

Finally, open a webpage, with no webpage defined, gives you:

image

I want to standardize these messages. Ids should be in quotes, and if no id is set it should use the empty string.

So these should all be

`Go to ""`
`Start an attempt for ""`
`Open ""`

I think a warning that an ID missing would be better, but we can do this first, then handle that.

maufcost:
Fixes #1511


jpeterson976 (approved):
LGTM!

zachberry (approved):
LGTM!

PR #1593, opened 2020-10-14
Converts text to a list if list-like formatting is detected

### LINKED ISSUE #1281: Automatically convert text chunks to lists when typing "* " or "1. "
When the user is in an empty line of a Text node and either types `* `, `1. ` or some other valid variation (like `A. `, `a. `, `c. `, `IV. `, etc) we should automatically covert that line into a List chunk. There's an old class called list-detector.js from the pre-Slate version of the editor which still exists in the codebase that was used to check if text was list-like. Either we could use that, or steal concepts / regexes / functions from that class as needed.

This would be helpful because I keep seeing examples of modules where the user is making a fake list, either by using text lines or manually inserting bullets, such as this:

image

jpeterson976:
Resolves #1281 

Hitting space after typing `*`, `1.`, or anything else list-like will cause the current line of text to be converted to a list. Any text that is on the same line will be copied into the list as well.

### DIFF
--- a/packages/obonode/obojobo-chunks-list/list-detector.js
+++ b/packages/obonode/obojobo-chunks-list/list-detector.js
@@ -1,8 +1,8 @@
 import ListStyles from './list-styles'
 
 const regexes = {
-	bulletListItem: /^[ \t]*(\*)[ \t]+([\s\S]*)/, // Interpert text such as ' * list item' as a bullet in a list
-	numericListItem: /^[ \t]*([0-9]+|[A-Za-z]|VIII|VII|VI|IV|IX|III|II|viii|vii|vi|iv|ix|iii|ii)\.[ \t]+$/, // Interpret text such as ' 1. list item' as a bullet in a list
+	bulletListItem: /^[ \t]*(\*)/, // Interpert text such as ' * list item' as a bullet in a list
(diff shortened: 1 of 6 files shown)


maufcost (approved):
Looking sharp! You just conquered Slate! The only thing I found, which I believe can be added to another issue is that, if an author would like to create an ordered list from a different starting point (like **8.** or **IV.**), the list if forced to start from 1. 
obo

zachberry (changes requested):
So happy this is getting built out! I did find another issue: When the list conversion happens some nodes below the newly created list get converted to text. Best seen in this example - The button below changes to simply a Text chunk with "Button":

zachberry (approved):
I did a little performance testing since I was worried this might be heavy, but convertIfList usually runs around 0.025ms, so it seems pretty negligible. Agree with @maufcost that being able to type "8. " to make a list start with 8 is desired, so I'll make that another feature to not stop this from making it in now. LGTM!

iturgeon (commented):
Functionality is flippin' awesome.  I had my head under a rock, didn't know about this coming and here it is 👍 

Tested: 
* 1.[space]
* 8.[space]
* 1.[tab] (doesn't work  - ok)
* III.[space] 
* [space][space]1.[space]
* 1[space].[space] (doesn't work - ok)
* a.[space] 
* all of the above INSIDE: code, heading, table, lists, figure (does nothing  - ok)

Seems pretty good to me.

PR #1600, opened 2020-10-22
Check lock before restoring document

### LINKED ISSUE #1576: Prevent users from reverting a document if it is currently being edited
Noticed while testing that it is possible to revert a document while it is being edited. The author doesn't know this happened and when they save the reverted document is overwritten.

Perhaps we need a check to make sure an open lock isn't in place before reverting.

ctcuff:
Before you restore a previous version of a document, that document is checked to see if anyone is editing it. You can still restore the document, but a warning dialog will appear before restoration. Just want to note that this dialog doesn't show the name of the user editing the document, that feature may have to be added at a later date as the username has to be retrieved from the `user_id` column in the `edit_locks` table.

Fixes #1576.


packages/app/obojobo-repository/shared/actions/dashboard-actions.js
```diff
@@ -279,5 +297,6 @@ module.exports = {
 	showModuleMore,
 	showVersionHistory,
 	restoreVersion,
-	importModuleFile
+	importModuleFile,
```
iturgeon, line 300: I don't see this being used in the diff, just verifying it's exported & added to Dashboard props is still needed?

packages/app/obojobo-repository/shared/actions/dashboard-actions.js
```diff
@@ -145,6 +156,12 @@ const restoreVersion = (draftId, versionId) => ({
 	promise: apiRestoreVersion(draftId, versionId)
 })
 
+const CHECK_MODULE_LOCK = 'CHECK_MODULE_LOCK'
+const checkModuleLock = draftId => ({
+	type: CHECK_MODULE_LOCK,
+	promise: apiGetModuleLock(draftId)
+})
+
```
iturgeon, line 164: 💯

iturgeon (approved):
Haven't run it, but the code all looks good.

Excellent job following the conventions etc. 👍

packages/app/obojobo-repository/shared/actions/dashboard-actions.js
ctcuff: Hmm, looks like it's still being passed and used in the Dashboard for uploads

https://github.com/ucfopen/Obojobo/blob/2626f5b26096987e97ce7f312b7a2da3a2dda6de/packages/app/obojobo-repository/shared/components/dashboard.jsx#L145

iturgeon: oops, that's my issue. I saw green + and managed not to see it was already in the previous version.  thanks for humoring me.

zachberry (approved):
LGTM!

Issue #1597, opened 2020-10-21
Splitting a styled heading can sometimes whitescreen the editor


zachberry:
I'm not sure what it is about this combo of inputs, but

1. Bold some part of an H1
2. In the middle of the bolded part, hit enter
3. Use the keyboard (not the mouse) to back up to the start of the bolded section
4. Hit enter again

Result: Whitescreen. It seems if you use the mouse to click at the start of the bolded section in step 3 the problem doesn't happen.

![2020-10-21 15 05 21](https://user-images.githubusercontent.com/73479/96770992-19eb7d00-13af-11eb-8ac6-b947fa976ebb.gif)

zachberry:
Another case - Selecting the start of a bolded section (here in a Button) and hitting enter:

![Uploading 2020-10-21 15.10.22.gif…]()

PR #1606, opened 2020-10-26
breakToText once again

### LINKED ISSUE #1599: Selecting a heading and lines of text then hitting enter errors out
Seen here:

vutoan1245:
Fixes #1599, #1597

1599 was due to `splitParent` cannot handle multiple nodes. I also further simplified it, not sure if it causes any issue.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
@@ -1,4 +1,4 @@
-import { Transforms, Point } from 'slate'
+import { Transforms } from 'slate'
 import { ReactEditor } from 'slate-react'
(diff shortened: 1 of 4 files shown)


ctcuff (changes requested):
Looks like selected a heading and lines of text no longer produces an error and styled text gets formatted into one line when split which is good! Also tested with figures, lists, code blocks, and HTML and everything seems to be working. One slight problem I noticed is that pressing enter at the beginning or end of a text node doesn't add a new line.

zachberry (dismissed):
I implemented some changes - I restored split-parent.js's functionality. Hopefully this isn't going around in circles, but this seems to include all of the fixes of @vutoan1245 's updates but also addresses @ctcuff 's finding that the new line was no longer working.

🤞

jpeterson976 (dismissed):
LGTM! Works great with my testing.

PR #1607, opened 2020-10-27
Added the option for authors to create lists from different starting indexes

### LINKED ISSUE #1596: Update convertIfList to set different start values
#1539 adds `convertIfList` which lets you type "1. " in a text node and it will convert that node to a list. This issue to make sure that if you type a non-1 character, like "8. ", "a. ", "X. ", etc, that the list will adopt the appropriate start value and list type. `convertIfList` uses `list-detector.js` which can report that type of information, we just need to connect that info to the ListStyles nodes of the list.

maufcost:
I added the option for authors to create lists from different starting indexes. So, if you type, for example, 3., 65., or 749., and hit space, the list should be created with the desired index. 

Note that roman numerals will also be translated to an ordered list with the proper number. So, if you type II. and hit enter, the list will be created with a starting index of 2.
Fixes #1596

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
@@ -70,7 +70,6 @@ Array [
             "deleteBackward": [Function],
             "deleteForward": [Function],
             "deleteFragment": [Function],
-            "getFragment": [Function],
             "history": Object {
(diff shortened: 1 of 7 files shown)


zachberry (dismissed):
This is awesome! Forgive me but I implemented an update which allows you to type "A." or "ii." or etc and get the expected result. It was more complicated than I expected, so I want to make sure I didn't break anything else with lists.

ctcuff (dismissed):
LGTM, tested a few numbered and lettered lists and they all work great!

PR #1608, opened 2020-10-28
Restrict allowed characters in id

### LINKED ISSUE #1601: Restrict input for IDs
If you input, for example, a `"` character in an ID field, and then revert to XML editing and then back to visual editing it can crash the editor. 

We need to determine exactly what are the allowed characters for the ids and then enforce those limitations.

jpeterson976:
Resolves #1601 

I found an older stale issue that suggested restricting to letters, numbers, hyphens (-), underscores (_), colons (:), and periods (.), so I display an error message if anything outside of that subset is entered into an id.


packages/app/obojobo-document-engine/__tests__/oboeditor/components/navigation/header.test.js
```diff
@@ -247,4 +247,58 @@ describe('Header', () => {
...
+		expect(component.instance().validateId('mock_id')).toBe(false)
+		expect(component.instance().validateId('mock:id')).toBe(false)
+		expect(component.instance().validateId('mock.id')).toBe(false)
+		expect(component.instance().validateId('mock+id')).toBe(true)
+		expect(component.instance().validateId('mock=id')).toBe(true)
+		expect(component.instance().validateId('abc123')).toBe(false)
+		expect(component.instance().validateId('ABC123')).toBe(false)
+		expect(component.instance().validateId('"mock-id"')).toBe(true)
+		expect(component.instance().validateId('/mock-id/')).toBe(true)
+		expect(component.instance().validateId('[mock-id]')).toBe(true)
```
zachberry, line 302: This is fine, but for future reference you can use `test.each` - The syntax is a bit strange but it works well for testing different inputs like you're doing - here's an example: https://github.com/ucfopen/Obojobo/blob/master/packages/app/obojobo-document-engine/__tests__/common/util/with-protocol.test.js

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/editor-component.js
```diff
@@ -50,6 +50,11 @@ class Node extends React.Component {
 		Transforms.select(this.props.editor, Editor.start(this.props.editor, targetPath))
 	}
 
+	validateId(newId) {
+		const validation = /[^a-z|A-Z|0-9|\-|_|:|.]/
+		return validation.test(newId)
+	}
```
zachberry, line 56: Duplicated code - I'm thinking move the validation function into it's own file in util, and import it where needed

zachberry (changes requested):

packages/app/obojobo-document-engine/src/scripts/oboeditor/util/validate-id.js
```diff
@@ -0,0 +1,6 @@
+const validateId = id => {
+	const validation = /[^a-z|A-Z|0-9|\-|_|:|.]/
+	return validation.test(id)
```
zachberry, line 3: I think this is a little confusing - `validateId` returns true if the given ID is *not* valid. Prefer to have this return true if the given id is valid.

zachberry (changes requested):
Looks great, I just have one final annoying code change and then I'll LGTM it

zachberry (approved):

PR #1611, opened 2020-11-18
update to lti in-memory nonce so it clears correctly

### LINKED ISSUE #1610: lti memory nonce doesn't clear correctly
The nonce isn't correctly clearing old nonce keys due to string and number addition issues.

iturgeon:
* updates dev_nonce_store so that it will properly expire old nonces
* adds max number of nonces to keep in memory
* updates dev express launches to use full timestamp as a nonce

fixes #1610


zachberry (changes requested):
As far as I could I tested this, mostly by calling it manually and logging out what's going on. I had a few questions and some suggestions.

If you try to call isNew with an expired timestamp the "`Timestamp is Expired" error will be logged out, however the nonce will remain in the used Map. Upon creating another nonce (calling isNew again) should call the `_clearExpiredNonces` method and delete the expired nonce at that point, however, do we want to allow it to remain in the map in the first place? As far as I can tell this is also an issue with the original code as well.

Suggestions:
1. Add a info log when a nonce is being deleted because the max size has been reached
2. Add a log when a nonce is being purged because its expired

If we get crashes from this code in the future those extra logs could be helpful to track down the reason.

iturgeon:
I'm inclined to resist adding the logs for clearing nonces we're going to get a ton of them.  I changed the message logged when cleaning to only log when something is cleaned, and to log the number that were cleaned.

zachberry (approved):

PR #1613, opened 2020-11-19
query timeouts added

iturgeon:
We had an outage where our AWS RDS database failed over to
its backup server.  Normally one would expect the db driver to lose
connection or at least somehow be aware of the outage.  It wasn't, and just
sat waiting for every db query to resolve.  No errors logged, no automatic
recovery.  The node proccesses needed to be terminated to re-establish
connections.

This commit adds query_timeouts to the configuration for pg-promise to ensure
that they will eventually time out.  We set them just below chromes usual request timeout
to maximize the time before timeout occurs.  This should give us a lot of headroom
for unforseen long-running queries and still produce error logs we can use to indicate
something has gone wrong.

Important: adds a required env var: DB_TIMEOUT, suggested value 29000

### DIFF
--- a/app.json
+++ b/app.json
@@ -14,6 +14,10 @@
 			"description": "PG-Promise SSL Configuration. https://github.com/vitaly-t/pg-promise/wiki/Connection-Syntax",
 			"value": "{\"rejectUnauthorized\":false}"
 		},
+		"DB_TIMEOUT": {
+			"description": "DB Query timeout in milliseconds https://github.com/brianc/node-postgres/pull/1760",
+			"value": "29000"
+		},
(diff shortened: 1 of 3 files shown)


zachberry (dismissed):
Not sure if I tested this the best way, but I artificially set the timeouts to 1ms and that caused the dashboard not to load (as expected). I tried different values to get (for example) starting an assessment attempt to fail, but I wasn't able to crash it. So that said, as far as I can tell it's working.

maufcost (approved):
Same here. I followed the same procedures as Zach, played with the timeouts, and everything seemed to work as expected. Good job!

PR #1614, opened 2020-11-19
adds fallback error page when res.unexpected isn't availible

iturgeon:
At least in development, I keep seeing several cases where express's catch all error handler is getting called and the res.unexpected function is undefined.

This patches that up a little for now - provides a fallback and makes sure the error that originally caused the problem doesn't get lost.  

Also adds a timestamp code to the error in the log and on the user's screen for the purpose of debugging them later.


maufcost (approved):
Everything looks good to me on Safari, Chrome, and Firefox. Great job! That is so much better than having that `res.unexpected` error pop up. Perhaps in the future, we could even help Obo developers see what is the cause of 500 errors when such errors are obvious and maybe easy to detect (e.g.: one hasn't started one's Docker environment yet).

zachberry (approved):

PR #1615, opened 2020-11-23
Adds prevent scroll toggle for focus:component event

### LINKED ISSUE #1570: Add preventScroll to the focus:component event Trigger/Action UI
#1123 adds `preventScroll` as a new option for the `focus:component` action. We need to update the Trigger/Action UI to add a toggle switch there. Also need to update the documentation (I'm holding off for now since not having this toggle UI could cause issues, and usually you won't need to enable this setting).

jpeterson976:
Resolves #1570

If both animate scroll and prevent scroll are turned on, then animate scroll will take over and prevent scroll will be ignored. I'll add this behavior to the documentation for now.


maufcost (approved):
Looking good on Safari, Chrome, and Firefox. Great job! There are a few times when the trigger modal is completely blurred, but I tested it before your changes and as it turns out, this "blur bug" was already there. I will talk to Zac about it and this will most probably be on another issue.

zachberry (changes requested):
(I merged in dev/16)

This is perfect, the only issue I found was weird - if you has scrollbars always set to show up and had enough triggers/actions to get them to show up in the dialog then the label for the combobox would move down a line. I cut out a word as a cheap way to get it to fit.

Anyway, blah blah, LGTM!

zachberry (approved):

maufcost (approved):
Everything looks good now. Good job, Jacob!

PR #1616, opened 2020-11-25
Max length for titles

### LINKED ISSUE #1251: Need a limit on title size
This is a minor issue, but we should cap the title length.

Absurd example below:

image

My thought is set the limit to 300 characters. That looks about like this (This is ~302 or so):

image

jpeterson976:
Fixes #1251 

I gave the title fields a maximum length of 300. I wanted to add a counter that shows how many characters have currently been entered, but I couldn't figure out a good way to do it so I only added the max length for now.


maufcost (approved):
Looking good in Chrome, Safari, and Firefox. Good job!

zachberry (approved):
LGTM!

PR #1617, opened 2020-11-30
Rectified how highest attempts are marked

### LINKED ISSUE #1504: Assessment Review - Highest attempt can be incorrectly marked
In the case of the tutorial - if I get a 100% on the first attempt and a 100% on the last attempt (which has scoring rules that reduce the 100% down to 95% ), both will be listed as the 'highest attempt'

Screen Shot 2020-09-10 at 12 22 55 PM
Screen Shot 2020-09-10 at 12 22 50 PM

maufcost:
Fixes #1504


ctcuff (dismissed):
LGTM, tested with a few assessments and they were all marked correctly!

zachberry (changes requested):
This can break with rubrics. If I submit an attempt score of 80% but the rubric defines an assessment score for 100% then the isHighestScore method will break because 80 !== 100. 

I think the answer is elsewhere - in full-review/index there's

```javascript
const highestAttempts = AssessmentUtil.getHighestAttemptsForModelByAttemptScore(...
```

I think we just need to change this to use the other method, `AssessmentUtil.getHighestAttemptsForModelByAssessmentScore`

maufcost:
After Zach's review, I kept finding more and more edge cases wherein the old solution broke. I changed a bit how I solved the problem, and now everything works on my end. Let me know if any changes should be made, and I'd be happy to.

jpeterson976 (approved):
LGTM! Works great with my testing.

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
```diff
@@ -36,13 +38,15 @@ class AssessmentReviewView extends React.Component {
...
-		const attemptReviewComponent = (attempt, isAHighestScoringNonNullAttempt) => {
+		const attemptReviewComponent = (attempt, isNonNullAttempt) => {
 			const date = new Date(attempt.completedAt)
 			const dateString = formatDate(date, "M/dd/yy 'at' h:mmaaaa")
 			const machineDateString = formatDate(date, "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")
 			const ariaDateString = formatDate(date, "MMMM Do yyyy 'at' h:mmaaaa")
 			const numCorrect = AssessmentUtil.getNumCorrect(attempt.result.questionScores)
 			const numPossibleCorrect = AssessmentUtil.getNumPossibleCorrect(attempt.result.questionScores)
+			const isAHighestScoringNonNullAttempt =
+				isNonNullAttempt && attempt.assessmentScore === highestAttemptScore
```
zachberry, line 49: I don't think this will work in every case, since assessmentScore and attemptScore are two different types of scores. attemptScore will always be a percentage of what was right or wrong, but assessmentScore could be anything based on the assessment rubric.

Instead I think calling on `AssessmentUtil.getHighestAttemptsForModelByAssessmentScore` _should_ work to get the highest scoring attempt(s)

zachberry (changes requested):

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
maufcost: Oh, I see that now. That makes sense.

maufcost:
I pulled the dev/16 snapshot (2020/2021) changes, but gh still shows that these snapshots are conflicting. I believe they are referencing my old conflicting commit to this PR. Let me know if you come across any errors or conflicts!

zachberry (changes requested):
Looks like the tests for packages/obonode/obojobo-sections-assessment/components/full-review/index.test.js are failing. As far as the snapshots, I think you may just need to re-run `yarn test -u` to update the snapshots?

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
```diff
@@ -138,6 +142,13 @@ class AssessmentReviewView extends React.Component {
 			)
 		})
 
+		// Deciding which is the highest attempt score
+		attempts.forEach(attempt => {
```
zachberry, line 146: I think this may not be necessary, since you have https://github.com/ucfopen/Obojobo/pull/1617/files#diff-91c983cfa647c49b998f8eb80445c3adbf590c93323345f792c6429e02d38bf3R30 - That line should give you an array of highest attempts, and may in the end be the only change needed in the PR.

zachberry (changes requested):

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
maufcost: That's true. I could have handled it just using _highestAttempts_. Working on it!

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
```diff
@@ -36,13 +36,19 @@ class AssessmentReviewView extends React.Component {
...
 
-		const attemptReviewComponent = (attempt, isAHighestScoringNonNullAttempt) => {
+		const attemptReviewComponent = (attempt, isNonNullAttempt) => {
 			const date = new Date(attempt.completedAt)
 			const dateString = formatDate(date, "M/dd/yy 'at' h:mmaaaa")
 			const machineDateString = formatDate(date, "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")
 			const ariaDateString = formatDate(date, "MMMM Do yyyy 'at' h:mmaaaa")
 			const numCorrect = AssessmentUtil.getNumCorrect(attempt.result.questionScores)
 			const numPossibleCorrect = AssessmentUtil.getNumPossibleCorrect(attempt.result.questionScores)
+			const highestAttempt = Math.max.apply(
```
zachberry, line 46: I don't think this is still necessary since `highestAttempts` from earlier still already has computed the highest attempt. That calculation is already handled by https://github.com/ucfopen/Obojobo/blob/f54e51cd9937359d3cc6c698f988f63673b9b208/packages/obonode/obojobo-sections-assessment/components/full-review/index.js#L150 and is what is passed in to the argument that was `isAHighestScoringNonNullAttempt`. I'm thinking revert the argument for this function back to `isAHighestScoringNonNullAttempt` or move the logic that I linked to to exist inside this function

zachberry (changes requested):

packages/obonode/obojobo-sections-assessment/components/full-review/index.js
maufcost: Jeez. That's simpler than I imagined. I hope it all works now! :)

ctcuff (dismissed):
LGTM, tested a few more assessments with different rubrics and it seems to be working!

PR #1620, opened 2020-12-03
Updated obo's favicon

### LINKED ISSUE #1121: Update favicon
We're using the older system favicon, would like to update this to match the new look

maufcost:
Fixes #1121


ctcuff (commented):
New icon looks great! Although it seems a bit large size-wise. The previous favicon was about 1.37 KB while this is about 99 KB. Not sure if it's too big of an issue, but it might be worth resizing down to 16x16 instead of 256x256.

maufcost:
That makes total sense. I will be doing that ASAP.

maufcost:
I managed to get the size down to 16 KB without losing too much quality. I tried with one version that got to 4 KB, but it got way too blurred. Let me know if you would like me to make the changes, and I would be glad to.

ctcuff (approved):
LGTM! Tried to compress it a little more but couldn't go down much further without losing some quality.

zachberry (changes requested):
Super happy this is getting put in! I wanted to compare the first one you made with the 16x16 on a MacBook retina display:

image

I think the 16x16 may be too small. Based on this - https://www.emergeinteractive.com/insights/detail/the-essentials-of-favicons/ - we probably need to go up to 32x32. I think that would give us a good balance of not-too-large but would still look sharp on more dense displays.

I also think, based on that article, we may need to pursue additional favicons based on their "A (mostly) complete list of all the known Favorite Icon Sizes" table, but I think we can spawn that off into another issue.

maufcost:
I decided to keep everything into this PR (including the additional favicon sizes according to Zac's article above) because this is a rather simple issue.

There are a simple observations that I would like to point out though:
- The favicons with the .ico extension in the latest commit of this PR range from 102KB-115KB. 
- The original .png favicons before I converted them into .ico were in the range of 12KB-16KB. I opted to convert them from .png to .ico because there are a few browsers that only accept .ico format for favicons. Let me know if you would like me to change them back to .png, and I will be glad to!
- I kept using serve-favicon to serve and cache the favicons, but I am not 100% if serve-favicon is loading the right icon on retina displays since I don't have one to test it with. Let me know if all the tests work out with you! If the favicons are not loading correctly yet, we might need to switch to serve-static. 

Here's Zac's article style guide that I followed for your reference:
favicon size guide

zachberry (changes requested):
Great work! I do think though pngs would be the way to go - they seem to be well supported now: https://caniuse.com/link-icon-png, and smaller file size would be a win.

Also, dev/16 was updated which fixes the 2020/2021 test issue, so I would undo the updates to the four snapshots and use dev/16's versions.

maufcost:
A couple of obs/takeaways from most of the research I did working on this issue (just in case anyone come across something similar in the future):
- As it turns out, serve-favicon does not serve/cache any favicons that are not .ico files. So, I could not use .png files as favicon despite its overall browser support. serve-static did not work as well because it is mainly used to serve/cache static files and not favicons.
- I managed to create a multilayered favicon **containing all the seven favicon sizes** (see the sizes on the table above) that will be loaded, served, and cached according to the user's device and browser. It is an .ico file and I managed to bring it down to only 41Kb (which is a huge plus) using GIMP.
- (Bonus) If anyone wants to convert png files to ico or in any way deal with png files on GIMP, ImageMagick, etc, make sure to _not_ compress png files beforehand. Compressing png files deletes part of the files' metadata, which these image processors need in order to correctly export files in the future (it took me a lot of time to figure this out :)).

I tested the committed multilayered favicon on retina displays, safari, chrome, and firefox. They are looking pretty good to me! Let me know if any changes need to be made!

chromefirefoxsafari

zachberry (approved):
Awesome!

jpeterson976 (approved):
Looks awesome on my retina display. Great work!

PR #1623, opened 2020-12-07
Fixes crash when resuming/finishing an updated assessment

### LINKED ISSUE #1587: Students unable to proceed if resuming an attempt for a module that was updated
I noticed this on my numeric branch, which may be the only branch experiencing the issue, but I believe it should be happening to master now. I think there's an issue with the following:

1. Student: Start an attempt on Draft A, version 1. Don't complete the attempt
2. Instructor: Update Draft A, which is now Draft A, version 2.
3. Student: Return to Draft A to resume your attempt. When submitting you get the following error:

image

Due to the following code in attempt-end.js:

```javascript
if (
	req.currentDocument.draftId !== attempt.draftId ||
	req.currentDocument.contentId !== attempt.draftContentId
) {
	throw Error('Cannot end an attempt for a different module')
}
```

The problem is that once you get this message you can never proceed, you'll always get this message.

My first thought on how to solve this - When saving a module we do a check for any in-progress attempts. If any are found we warn the author that by changing this module they'll invalidate X attempts. If they agree, then any in-progress attempts are deleted. Should a student return to the module they won't be forced to resume the attempt and will have to start the attempt over. This would prevent them from getting stuck.

ctcuff:
When an assessment is updated while a student is currently taking it, attempting to submit that assessment invalidates it and prompts the user to start a new attempt.

When a user returns to an assessment that was updated while they were away, that attempt gets invalidated and they'll see the same "Unfinished Attempt" dialog but clicking resume starts a new attempt.

Invalidation happens on the server-side and restarting the attempt happens on the client-side in the `Assessment Store`.

**NOTE**: Restarting an attempt after it was invalidated does **not** change the total amount of attempts remaining or taken. For example, starting an assessment with 1/3 attempts then restarting after invalidation starts the assessment again with 1/3 attempts.

**IMPORTANT**: Invalidating an attempt adds an `invalid` key to the `state` json column in the `attempts` table. Any attempt that's invalid will have this key set to `true`. By default, attempts that aren't invalid don't have this key. This allows us to ignore and filter out any invalid attempts without actually deleting those attempts. **However**, when restarting an attempt, it's assumed that the assessment ID doesn't change. This is important since that ID is what's actually needed to restart an attempt. Because of that, this solution only works with an unchanged assessment ID.

Fixes #1587.


packages/obonode/obojobo-sections-assessment/server/express.test.js
```diff
@@ -193,9 +193,12 @@ describe('server/express', () => {
 	test('POST /api/assessments/attempt/mock-attempt-id/resume errors', async () => {
 		expect.hasAssertions()
 		const mockReturnValue = {}
+		const errorString = 'Cannot resume an attempt for a different module'
```
zachberry, line 196: There's a lot of inline constants, possible to move these out into their own file that can be shared between the different assessment server files?

zachberry (changes requested):
This is working fantastically.

I think it's okay that if the assessment ID changes it breaks. When restarting you get an unexpected DB error, which could be better. But if you refresh the page you aren't stuck and are able to start a new assessment.

I have a thought - should the invalid flag live as its own column? I could see is_invalid being useful. Right now leaning towards that, but I'm open to talk with you and the team.

I also have a change on the wording of the updated module dialog - I didn't go ahead and change it because I wanted to get feedback on it before I did. I'm attaching it below:

Screen Shot 2021-01-22 at 12 51 14 PM

ctcuff:
@zachberry I like the dialog suggestion so I'm ok with updating that 👍. As for the DB error, I'm not sure what the best way to handle this is since the assessment depends on this ID (I think???) but that may be a rare case and worth opening a new issue for. Although, I think it could at least be handled a little more gracefully in this issue perhaps. I'm also open to moving the `invalid` flag out of the state column as long as it doesn't affect other assessments but I do want to get @iturgeon's opinion on that though.

packages/obonode/obojobo-sections-assessment/server/express.test.js
ctcuff: That'd definitely simplify things! Do you have any suggestions on where I could move these?

ctcuff:
Mini update: The latest commit includes [`error-constants.js`](https://github.com/ucfopen/Obojobo/blob/b27a77a840c16e4cdae11700386fbe69bba60e14/packages/obonode/obojobo-sections-assessment/server/error-constants.js). As of now, this file only contains constants related to errors returned by the assessment server when an attempt is invalid. These constants are also used in `assessment-store.js`. Right now, it's placed under `obojobo-sections-assessment/server/error-constants.js` but I'm open to moving it around/modifying what constants are actually stored in this file.

jpeterson976 (approved):
This is great, LGTM!

iturgeon:
A note for future us - mind the notes above about changing the attempt id -   but if you do test changing the attempt id make sure you update the attempt start buttons actions to associate with the new assessment id.  currently the error looks the same to the client "unknown db error".  There is already an issue about this here https://github.com/ucfopen/Obojobo/pull/1624

PR #1605, opened 2020-10-26
Clearer message for assessment ID error

ctcuff:
Attempting to start an assessment with a trigger using an assessment ID that couldn't be found now shows an error modal:

Screen Shot 2020-10-26 at 5 15 00 PM

Fixes #1573

Issue #1573, opened 2020-10-05
Starting an attempt for an assessment that can't be found shows Unexpected DB Error


zachberry:
1. Modify an example document by changing the assessment id
2. Preview the module and try to click the button to start the assessment

Result: "Unexpected DB Error"

We should try to make this error more clear in what's actually happening. It could help the author figure out the issue.

PR #1624, opened 2020-12-08
Fixed invalid assessment id error

### LINKED ISSUE #1621: More helpful error message when starting an attempt with a bad id
Found in production with a user - If you wire up an ActionButton to start an assessment but give it an assessmentId that doesn't exist in the document the app will ask the server to start that assessment and an unhelpful "Unexpected DB Error" will be returned. Instead, when clicking on action buttons we should throw an error if the ID referenced doesn't exist in the system.

A further improvement would be to warn authors in the editor if an ID is invalid with an error message. I think we should do both, but feel free to spawn that into another issue if needed.

maufcost:
I thought of a couple different new error messages and I went with: "This assessment id does not exist".

A couple of other message ideas I had:
- _"This assessment id is invalid"_ (I think this is not really helpful since it may confuse authors on what "invalid" really means).
- _"Obojobo could not find this assessment."_

I think it would also be helpful for authors to troubleshoot this kind of error if we could add a small subtext under that message saying something like: _"Did you input a correct assessment id for this button?"_

Also: I will be adding that idea about warning authors about invalid assessment ids on another new issue since it requires a few different changes and implement that there.

Let me know if anything should be changed.
Fixes #1621

### DIFF
--- a/packages/obonode/obojobo-sections-assessment/server/attempt-start.js
+++ b/packages/obonode/obojobo-sections-assessment/server/attempt-start.js
@@ -10,6 +10,7 @@ const ACTION_ASSESSMENT_SEND_TO_ASSESSMENT = 'ObojoboDraft.Sections.Assessment:s
 const ERROR_ATTEMPT_LIMIT_REACHED = 'Attempt limit reached'
 const ERROR_UNEXPECTED_DB_ERROR = 'Unexpected DB error'
 const ERROR_IMPORT_USED = 'Import score has already been used'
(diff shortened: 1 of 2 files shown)


jpeterson976 (approved):
This looks good to me. My own idea for the error message is some variation on "This assessment could not be found", though I like the current one and how it specifies the ID is the issue so I'm giving it my approval.

zachberry:
Sorry for the mixup :( . Closing in favor of #1605. I had created an issue on this problem - #1573 - and had forgotten that existed when I created #1621.

PR #1629, opened 2020-12-11
Fixes invalid start value crash

### LINKED ISSUE #1577: An invalid "start" value for a module breaks the module
If you give your module a start value, like `<Module start="page-that-does-not-exist" ...>` the viewer will throw errors and won't display pages. Additionally, the Editor should be modified so that it doesn't listen for the start value - right now if you give the editor a start value and refresh the page it will go to that page at first. This is (for me) unexpected, and it also shows a blank page if you have an invalid start value.

ctcuff:
* The editor no longer listens to the starting value, it now always defaults to the first page in the module
* If the starting ID isn't found, the viewer defaults to the first page in the module

Fixes #1577


zachberry (approved):
Perfect

jpeterson976 (approved):
Works great!

PR #1631, opened 2020-12-14
Clear selection on page change

### LINKED ISSUE #1618: Clear selection when changing pages
I noticed this today - It seems that text selection can be maintained on page changes in the viewer

ctcuff:
Any text highlighted on a page now gets cleared when navigating to different pages using the nav button at the top/bottom of modules. For this issue, I've added a `clearSelection()` function to the `nav-util` but I'm open to moving it to a different file if appropriate.

Fixes #1618


maufcost (approved):
LGTM! I tested with a myriad of module numbers, and they worked great on my end. Good job!

zachberry (changes requested):
This clears the selection when using the top and bottom buttons but doesn't when using the left-hand nav menu. Also I'm weary of doing any DOM manipulation in the *-util classes (not that we're perfect about that).

Instead, I'm thinking add something to viewer-app's componentDidUpdate method. There's  a method that gets called there - `scrollToTopIfNavTargetChanging` - We could add a similar method where that gets called - `clearSelectionIfNavTargetChanging`. (Or combine both into `scrollToTopAndClearSelectionIfNavTargetChanging`, but nobody seems to like my super-long function names)

zachberry (approved):
(I merged in dev/16)

Awesome, approved!

PR #1633, opened 2020-12-16
Fixes viewer crash when loading assessment with katex

### LINKED ISSUE #1604: Katex content in assessment only can crash viewer
If you include katex content in an assessment quiz (but nowhere else in the module) the assessment can crash the viewer. This is because window.katex is not loaded since inject-katex-if-needed won't pick up on katex content inside assessment questions (they aren't loaded until an attempt is started).

Easiest fix is to call inject-katex-if-needed when loading an attempt.

ctcuff:
Katex is injected when starting/resuming an assessment since it might not be loaded when the document is first rendered.

Fixes #1604

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/common/util/inject-katex-if-needed.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/common/util/inject-katex-if-needed.test.js
@@ -24,6 +24,12 @@ describe('Inject Katex if Needed Util', () => {
 		expect(global.window).not.toHaveProperty('katex')
 	})
 
+	test('skips loading if katex is already defined', async () => {
+		global.window.katex = 'mockObjectWithKatexLoaded'
+		await injectKatexIfNeeded({ value: { text: 'latex' } })
+		expect(global.window.katex).toEqual('mockObjectWithKatexLoaded')
+	})
+
 	test('handles error when katex loading times out', async () => {
 		expect.assertions(2)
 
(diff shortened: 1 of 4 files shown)


maufcost (commented):
The viewer is not crashing anymore, but I don't think Katex equations are loading correctly on the assessment (I might be doing something incorrectly). Could you try with `\overleftrightarrow{AB}`, for example, and let me know if it works on your end?

Latex Issue

maufcost (changes requested):
As it turns out, I had not switched branches....
I tried with a simple latex command like `\overleftrightarrow{AB}` and it renders **perfectly** on the page, but as soon as I submit a response to a question, the viewer crashes again :(

ctcuff:
> As it turns out, I had not switched branches....
> I tried with a simple latex command like `\overleftrightarrow{AB}` and it renders **perfectly** on the page, but as soon as I submit a response to a question, the viewer crashes again :(

Good catch! Turns out I forgot to inject katex when rendering the assessment review. Should _hopefully_ be fixed now.

maufcost (dismissed):
LGTM! Great job!

zachberry (changes requested):
This is working great but I'd like one minor optimization - On multiple attempts injectKatexIfNeeded will run multiple times and potentially inject katex multiple times. injectKatextIfNeeded could look for the presence of `window.katex` and finish right away if it already exists.

zachberry (approved):
LGTM!

PR #1639, opened 2021-01-04
Docker in Production

iturgeon:
Purpose: create a docker image that people can use in production. Reduce & simplify our deploy processes.

* moves some processes built into our private obojobo-deploy repository into the main repo
* Creates a production ready docker obojobo application and nginx webserver
* all workspace dependencies on other obojobo npm modules moved to peer dependencies to allow installing to work
* uses a small wrapper package in the docker image to add aws secret loading & pm2 etc (serves as a nice example of customization too)


docker/config/nginx/nginx-prod-with-load-balancer.conf
```diff
@@ -0,0 +1,108 @@
...
+            proxy_set_header X-NginX-Proxy true;
+            #proxy_set_header Upgrade $http_upgrade; # needed for websockets
+            #proxy_set_header Connection "upgrade"; # needed for websockets
+
+            proxy_pass http://obojobo_upstream/;
+            proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_read_timeout 240s;
+
+            #proxy_cache one;
```
zachberry, line 104: Intentionally commented lines here? If so, need a comment I think as to why

zachberry (changes requested):
Still got a lot to learn with Docker so I can't give you much in terms of docker review, but I did a casual run through trying to see if I could find anything that might be an issue

docker/config/nginx/nginx-prod-with-load-balancer.conf
iturgeon: Removed for now, it's for caching responses, but it'll take some testing & tuning that aren't on the books right now

zachberry (approved):

iturgeon:
Sorry, I made a minor readme update after Zach approved, won't affect functionality.

zachberry (approved):

PR #1642, opened 2021-01-05
fix year dependent snapshots for 2020

iturgeon:
So the year flipped from 2020 to 2021 and there are a couple of snapshots that have the current year shown in the footer in them.

This pr mocks the date to a constant value that wont change.


jpeterson976 (approved):
LGTM!

ctcuff (approved):
LGTM!

PR #1643, opened 2021-01-06
Makes new assessments lock navigation by default

### LINKED ISSUE #1632: Make "Lock Navigation During Attempts" enabled by default
Sudden thought, but I think most instructors would assume assessment would be locked down, and would want this to happen by default. I think newly created modules could have the triggers/actions necessary to enable "Lock Navigation During Attempts".

jpeterson976:
Resolves #1632


maufcost (dismissed):
LGTM! Good job!

packages/app/obojobo-document-engine/documents/empty.xml
```diff
@@ -90,6 +90,22 @@
 					</Page>
 				</scoreAction>
 			</scoreActions>
+			<triggers>
```
ctcuff, line 93: Minor style suggestion, not an expert on this but you could remove the empty value tag to simplify this a bit:
```xml
<triggers>
	<trigger type="onStartAttempt">
		<actions>
			<action type="nav:lock" />
		</actions>
	</trigger>
	<trigger type="onEndAttempt">
		<actions>
			<action type="nav:unlock" />
		</actions>
	</trigger>
</triggers>
```

ctcuff (approved):
Small suggestion below but besides that, everything looks good!

zachberry (changes requested):
(I merged in dev/16 to fix the snapshot timestamp issue)

I like @ctcuff 's suggestion - once that's addressed I'll give this the stamp of "LGTM"

PR #1645, opened 2021-01-12
Fixes header menu focus

### LINKED ISSUE #1528: Focus rectangle in the Header menu stays stuck on the last item when using keyboard
Originally reported in #1490 

If you use the arrow keys when the heading menu is open then close the menu, the next time you open the menu the focus rectangle is still at the last selected item. This can be confusing because it makes it look like the system thinks the heading you are editing is at the wrong level.

Example:

When opening the menu back up the focus rectangle should either be at the correct item or be on the first item (Normal text)

ctcuff:
The focus rectangle in the menu now focuses on the current paragraph style when opened. If multiple nodes are selected when the menu is opened, it defaults to focusing on the first item.

Fixes #1528


zachberry (approved):
I merged in dev/16 to fix the test timestamp issue. But yeah this looks great!

maufcost (approved):
This is awesome! Great job. LGTM.

PR #1646, opened 2021-01-19
Stronger tests for breakToText

### LINKED ISSUE #1484: Stronger tests for KeyDownUtil.breakToText
These tests simply ensure that various methods are called - Need to test to make sure the method converts the data structures as expected (Heading becomes a Heading and a Text Chunk, for example)

jpeterson976:
Resolves #1484 

I added an additional check to make sure the node fragment that's grabbed is the correct type before it's converted, and an additional test that explicitly tests for splitting in the middle of a node. Let me know if there's anything else I missed that should be tested.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
@@ -1,4 +1,4 @@
-import { Transforms } from 'slate'
+import { Transforms, Node } from 'slate'
 import { ReactEditor } from 'slate-react'
 jest.mock('slate-react')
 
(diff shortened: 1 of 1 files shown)


zachberry (changes requested):
There might be some issues here with #1606 which changes some of the breakToText functionality (I'm sorry!!!)

packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
```diff
@@ -348,11 +352,53 @@ describe('KeyDown Util', () => {
...
+
+		KeyDownUtil.breakToText(event, editor, [editor.children[0], [0]], true)
+
+		expect(event.preventDefault).toHaveBeenCalled()
+		expect(Transforms.insertNodes).toHaveBeenCalled()
+
+		// make sure the correct node fragment is taken
+		const fragmentNode = Node.fragment.mock.results[0].value[0]
+		expect(fragmentNode).toHaveProperty('type', HEADING_NODE)
+		expect(fragmentNode.children[0]).toEqual({ text: 'Text' })
```
zachberry, line 393: Should this have the text of `some` since it split `someText`?

zachberry (changes requested):

packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
jpeterson976: The fragment is the part of the node that's being broken off, so it'd have `Text` in this case. The inserted node uses the text from the fragment node, so this is just confirming it has the correct source type and text.

zachberry (approved):

maufcost (approved):
It all looks good to me. Good job, Jacob!

PR #1656, opened 2021-02-02
Creates assessment:attemptInvalided event

### LINKED ISSUE #1655: Add assessment:attemptInvalidated event
Due to #1623, we need an event for when assessments are invalidated

ctcuff:
`insertEvent` is called after an assessment is invalidated. Fixes #1655


zachberry:
I made a tweak on this - I changed it so that it will only invalidate the attempt if it wasn't previously invalidated. That means only one event will be sent even if the student keeps attempting to submit the bad attempt

maufcost (approved):
This looks pretty good to me. Amazing job, Cam and Zac!

I tested with a bunch of assessment questions and rubrics, including updating questions and rubrics while a "student" (me) was taking the assessment. The modal saying that new updates were made to the assessment is so much better and worked with all the tests cases I used.

The only thing is that now obo shows the old error on the console, but it doesn't really hinder any new functionalities, so I believe this is good to go:
Screen Shot 2021-02-04 at 10 20 15 AM

iturgeon (changes requested):
Just a few comments code suggestions

iturgeon (approved):

PR #1695, opened 2021-02-25
Simplify the JSON->XML and Slate -> Obo text conversions

### LINKED ISSUE #1694: JSON->XML parsing and Slate->Obo text conversion adds unnecessary styles.
Both of these processes are attempting to replicate the data structures of StyleableText, but that class has methods to handle adding and normalizing styles. Both should be using StyleableText and its features to produce simpler XML and JSON.

zachberry:
Fixes #1694 

* In Visual Editor’s TextUtil.slateToOboText method, the conversion from Slate text to Obo text is done by using the StyleableText class, which already handles many of the transformations slateToOboText is replicating.
* In the JSON parser’s text-group-parser, simplifies the conversion to XML. Also enforces normalizeStyles, making sure the resulting XML does not include unnecessary style elements

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/util/text-util.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/util/text-util.test.js
@@ -167,6 +167,11 @@ describe('TextUtil', () => {
 					type: 'a',
 					href: 'mockotherURL',
 					children: [{ text: 'atted' }]
+				},
+				{
+					text: 'abc',
+					_latex: true,
+					monospace: true
 				}
 			]
 		}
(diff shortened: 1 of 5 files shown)


zachberry:
These changes touch some key core code:

1. StyleableText and related methods (these are used to render text in the Viewer)
2. Conversion between the StyleableText format and the slate data model for text (used to render text in the Editor)
3. Conversion for text from JSON -> XML.

Therefore, testing this would involve making various blocks of formatted text throughout the different chunk types and

1. Ensuring that changing the editing interface between Visual, XML and JSON options doesn't result in any data loss.
2. Ensuring that the preview of a document always matches the display of the editor, and visa-versa (Text is formatted the same, no text is missing)

maufcost (dismissed):
LGTM!

I created a couple of text chunks (some of them separated by other chunk types) and applied some different styles to them, including styles that involved selecting multiple text chunks (or some part of them). They all kept their style from the visual editor -> viewer.  I also made some similar changes on the JSON and XML editors, and all the changes were consistent on the viewer as well. Great job, Zac!

jpeterson976 (approved):
LGTM! My testing is consistent with Mauricio's, all three editors look to be applying styles correctly.

PR #1311, opened 2020-05-11
Improve the Toolbar

vutoan1245:
Resolves #1291 

Preview:
Toolbar

PR #1704, opened 2021-03-09
Improve toolbar update

### LINKED ISSUE #1291: Clicking on a toolbar menu item leaves it open even when another opens

jpeterson976:
Resolves #1291 

This is an update to #1311 since a lot had changed in the code since it was opened, so it's just for @iturgeon  to reference when he works his git magic so the old code I based this on isn't lost to the void.


iturgeon:
I tried, but it looks like getting Toan's changes into history is going to be more trouble than it's worth.  😢

zachberry:
Closing in favor of #1755

PR #1719, opened 2021-03-17
Fixed escape issues when opening modals

### LINKED ISSUE #1377: using escape to exit iframe (and other chunks) leaves editor disabled

maufcost:
Fixes #1377


maufcost:
I am going to close this PR since I just realized I worked on the issue based off obo's main repo and not my forked one (maufcost/Obojobo). I will submit it again from there.

iturgeon:
Ok, let's delete the branch on the main repo when it's safe to do so @maufcost

PR #1723, opened 2021-03-18
Fixed escape issues when closing modals

### LINKED ISSUE #1377: using escape to exit iframe (and other chunks) leaves editor disabled

maufcost:
I dealt with some edge cases (including Zac's on Ian's old PR above). There are also cases where modals, such as the iframe-properties-modal adds a scrollbar on the right - thus making the X button appear on top of it - so I fixed that as well. 

All the other modals are looking good to me as far as my testing goes. Let me know what I should change, and I'd be happy to.

Fixes #1377


jpeterson976 (commented):
Works great! I checked all the modals I could think of just to be sure, and I think the only one with a scrollbar that doesn't have the extra offset on the close button is the mod properties modal for assessments. All the rest either don't have a scrollbar or have the offset.

maufcost:
The modal case above has been fixed! Cases such as the mod properties modal for assessment only _really_ need the extra offset when we add enough mod properties, leading to the addition of the right scrollbar.

For now, I left the x button on this case with the offset before and after the right scrollbar, and it looks alright to me. Let me know if you would like me to change anything else.

iturgeon:
How's this relate to https://github.com/ucfopen/Obojobo/issues/1377

maufcost:
Since Ian left, I believe his #1378 is probably gonna go stale soon. I applied the changes that Zac requested on that PR to this one. Let me know if you would like me to make any changes.

Also, for us and future generations, I created a wiki page with the list of all the modals in Obojobo (it may facilitate tests and reduce the number of back-and-forth PRs and change requests): https://github.com/maufcost/Obojobo/wiki/Obojobo-Modals

jpeterson976 (dismissed):
LGTM! The addition of the wiki page is extremely useful, that was a great idea. I tested all the modals listed on there and they all work and display as expected.

jpeterson976 (approved):
Re-tested after merging in dev/25 to an untracked branch and it still worked great. LGTM!

zachberry (changes requested):
The resume dialog is the problem child - it's showing a (X) button when it shouldn't. Thankfully clicking it and hitting ESC does nothing.

image

PR #1731, opened 2021-03-19
Updates Materia captions so they no longer duplicate the node on Enter

### LINKED ISSUE #1728: Hitting enter in the caption for a Materia node duplicates the node
Instead this should work like Figure. We need to add this to the editor-registration file:

```js
onKeyDown(entry, editor, event) {
	switch (event.key) {
		case 'Enter':
			return KeyDownUtil.breakToText(event, editor, entry)
	}
},
```

zachberry:
Fixes #1728


iturgeon (approved):
Tested in chrome, hitting enter mid caption and at end of caption

PR #1738, opened 2021-03-22
enable optional node config

iturgeon:
Flips environment variable OBO_DISABLE_NODES to OBO_OPTIONAL_NODES
    
* adds environment variable option to enable optional nodes
* OBO_OPTIONAL_NODES can now be set with a comma separated list to enable optional nodes
* optional nodes are identified with isOptional: true in their manifest


iturgeon:
new machine, set up my remotes all weird.. fixing..

PR #1745, opened 2021-03-24
Max-width option for figure captions

### LINKED ISSUE #1707: Figure option: Caption length
Sometimes, the fact that the caption of a figure is at the same width as the image can be harmful, as shown in this example:

image

Since the caption is so large it results in a weird-looking page. I'm thinking we add a toggle option that would let a caption either fit to the same width as the image, or, 100% width, as the mockup below is shown (using an H5 instead of the caption):

image

jpeterson976:
Resolves #1707 

Added a select box that lets you choose between image-width and max-width for figure captions. The default is image width, since that's the current behavior.


zachberry (changes requested):
I didn't review the code but I did test it in the browser. I see it working in the editor but in doesn't seem to work in the viewer. Additionally it doesn't seem to be working for custom sizes (I tried 100x100)

packages/obonode/obojobo-chunks-figure/image-properties-modal.test.js
```diff
@@ -201,7 +201,23 @@ describe('Image Properties Modal', () => {
...
+		const component = mount(
+			<ImageProperties
+				content={{ url: 'mock_url', captionWidth: 'image-width' }}
+				onConfirm={jest.fn}
+			/>
+		)
+
+		// Simulate changing "Caption Width"
+		component
+			.find('#obojobo-draft--chunks--figure--caption-width')
```
iturgeon, line 214: 👍

packages/obonode/obojobo-chunks-figure/__snapshots__/editor-component.test.js.snap
```diff
@@ -58,14 +58,15 @@ exports[`Figure Editor Node Figure component 1`] = `
...
+exports[`Figure Editor Node Figure component handles tabbing 1`] = `"<div><div class=\\"obojobo-draft--chunks--figure viewer undefined  is-not-selected\\"><div contenteditable=\\"false\\" class=\\"accessibility-warning\\" style=\\"user-select: none;\\">Accessibility Warning: No Alt Text!</div><figure class=\\"container\\"><div class=\\"figure-box   is-not-selected\\" contenteditable=\\"false\\"><div class=\\"obojobo-draft--components--button is-not-dangerous align-center delete-button\\" contenteditable=\\"false\\"><button class=\\"button\\" tabindex=\\"-1\\" contenteditable=\\"false\\">×</button></div><div class=\\"image-toolbar\\"><div class=\\"obojobo-draft--components--button is-not-dangerous align-center properties-button\\" contenteditable=\\"false\\"><button class=\\"button\\" tabindex=\\"-1\\" contenteditable=\\"false\\">Image Properties</button></div></div><div class=\\"img-placeholder\\"></div></div><figcaption class=\\"align-center is-undefined\\"></figcaption></figure></div></div>"`;
```
iturgeon, line 69: `is-undefined` seems undesirable here.  Does this indicate we need a default value in the editor component?

Keep in mind that previously created Figure chunks will render `is-undefined` like this sample

iturgeon (changes requested):
It looks like the viewer defaults to image-width for figures missing the caption width setting, the only thing not handling it is the style in the editor that results in `is-undefined` getting rendered to the dom.  We should either default to image-width or just avoid displaying the is-undefined at all

I expanded use of the image-width constants to everywhere they show up in this chunk.

packages/obonode/obojobo-chunks-figure/adapter.js
```diff
@@ -21,6 +23,7 @@ const Adapter = {
 		model.setStateProp('width', null, p => parseInt(p, 10))
 		model.setStateProp('height', null, p => parseInt(p, 10))
 		model.setStateProp('alt', null)
+		model.setStateProp('captionWidth', ImageCaptionWidthTypes.IMAGE_WIDTH)
```
zachberry, line 26: This should lower-case the attribute and ensure that it only be one of the two acceptable values:

```suggestion
		model.setStateProp('captionWidth', ImageCaptionWidthTypes.IMAGE_WIDTH, p => p.toLowerCase(), [ImageCaptionWidthTypes.IMAGE_WIDTH, ImageCaptionWidthTypes.MAX_WIDTH])
```

packages/obonode/obojobo-chunks-figure/viewer-component.js
```diff
@@ -24,16 +30,18 @@ const Figure = props => {
...
 				className={`obojobo-draft--chunks--figure viewer ${props.model.modelState.size}`}
 			>
 				<div className="container">
-					<figure unselectable="on" style={customStyle}>
-						<Image chunk={props.model} />
+					<figure unselectable="on">
+						<Image chunk={props.model} style={customStyle} />
 						{props.model.modelState.textGroup.first.text.length > 0 ? (
-							<figcaption>
+							<figcaption className={`is-${captionWidth}`} style={captionStyle}>
```
zachberry, line 44: I think we need to change this to `is-caption-width-${captionWidth}`

zachberry (changes requested):
Looks great! I just have a few minor suggestions

zachberry (approved):

PR #1753, opened 2021-03-30
Fixed iframe zoom property not being updated

### LINKED ISSUE #1737: Iframe initial zoom values not being respected in specific cases
The issue is due to the fact that the IFrame constructor is calling MediaUtil.setDefaultZoom. If you happen to have a page of IFrames followed by a page of IFrames, and switch between pages, the second page of IFrames won't call MediaUtil.setDefaultZoom.

Instead, we should move the MediaUtil.setDefaultZoom call into componentDidMount (will need to test to make sure this still works with IFrames with autoload=true)

maufcost:
I found out what was going on, and I fixed it by using the viewer's iframe's componentDidUpdate method.

Just a couple of more observations: 
- Take a look at the drawing I did on #1737. When I switch from page 1 to page 2, React does not unmount the iframe 1 and 2. It re-uses both components and now updates them with the new zoom data with my new updates + it mounts the extra iframe (iframe 5) as expected.
- It will **only** unmount **all** the iframes in a page, if I switch from a page with no iframes, such as page 3 back to a page with iframe(s).

As always, let me know if you would like me to make any change, and I'd be happy to.

Fixes #1737


maufcost:
Everything seems to be working fine now. Using `mediaState` really makes a lot more sense here. I didn't manage to break it again. Let me know if anyone finds some more edge cases.

iturgeon:
Testing out your code, it looks like it's functioning to me... however, I think I found the source of the issue:

It looks like it's because Page gives each child a key based on its index in the Page.  this causes the react component instance to be reused.  So if you have an iframe on page 1 that is the 3rd item in the page, it'll be the same object when you switch to page 2 if there's an iframe that is the 3rd item on that page... causing the internal state and variables to bleed from one to the other.

I tested fixing Page's keys and the problem completely goes away.  Upside: simple fix, downside: have to undo some work.

maufcost (commented):
Changing the key of the child elements to be their `uuid` works great and correctly forces unmounting and mounting again of components.

I tested it on a couple of iframes and pages, and everything looks good. Good job.

PR #1757, opened 2021-04-05
Fix non-failing test warnings

### LINKED ISSUE #1667: Fix non-failing test issues in the Dev/16 release
There's a few error or warning logs being output in the tests for dev/16 - These don't trigger a failed test but do output when running the tests. I did a review and these all seem to be issues with the writing of the tests and not indicative of an actual code issue, so we're going to leave them as is for dev/16.

To better see them, run `yarn test:ci`

* packages/app/obojobo-repository/shared/components/dashboard.test.js: The `'"New Module" and "Upload..." buttons call functions appropriately'` test isn't correctly mocking useState as it expects, resulting in several `wrap in act()` errors
* Error: Mocked FS doesn't have mock file contents for .../packages/app/obojobo-express/server/config/db.json
* Error: Missing config environment for "default" and "test" for .../packages/app/obojobo-express/server/config/db.json
* Warning: Received `false` for a non-boolean attribute `crossOrigin`.
* Assessment Store: Some jest spies for "startAttemptWithImportScoreOption" need to mock implementation: `jest.spyOn(AssessmentStore, 'startAttemptWithImportScoreOption').mockImplementation(jest.fn())`

jpeterson976:
Resolves #1667 

Ian got most of the ones listed on the main issue, this fixes the last assessment store one. I ended up just doing the `.mockImplementation(jest.fn())` fix instead of something with `.mockResolvedValue()` since the resolved value isn't used in the tests producing the warnings at all.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/Viewer/stores/assessment-store.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/Viewer/stores/assessment-store.test.js
@@ -1504,7 +1504,7 @@ describe('AssessmentStore', () => {
 		AssessmentAPI.resumeAttempt.mockResolvedValueOnce(mockResumeAttemptResponse)
 
 		jest.spyOn(AssessmentStore, 'updateStateAfterStartAttempt')
-		jest.spyOn(AssessmentStore, 'startAttemptWithImportScoreOption')
+		jest.spyOn(AssessmentStore, 'startAttemptWithImportScoreOption').mockImplementation(jest.fn())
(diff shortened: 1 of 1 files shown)


iturgeon (approved):
While `.mockImplementation(jest.fn())` works because of the reasons you mentioned AND a lot of the existing tests in this same test use this same approach.

I'd argue that mocking async/Promise methods with non async functions is incongruent with the actual implementation.

I'll approve cause it's working and consistent w/ the test, but think we should update these wholesale where they don't match async implementation.

zachberry (approved):
In regards to @iturgeon 's comments, #1408 moves this method elsewhere, so I think it's okay as is for now

PR #1766, opened 2021-04-07
Figure caption width in tutorial

### LINKED ISSUE #1762: Update visual editor tutorial to cover new caption width options for figures
Needed now that #1745 is in

jpeterson976:
Fixes #1762 

I may have added a little too much to the new tutorial section, so let me know if I should cut it back at all.

I also changed `max-width` to `text-width` since that's a little more accurate, and made a minor CSS change for text width captions to better account for medium and large sizes.


iturgeon:
The only argument I have against Text-width is when using captions on large images that extend past the text width.  The image below shows the text width vs caption width under a large figure

Screen Shot 2021-04-07 at 4 53 59 PM

zachberry:
@jpeterson976 @iturgeon Both good points. `max-width` is used for Iframes for sizing which means beyond text width, so using `max-width` for figure width is probably a mistake. But there is the weird nuance that Ian brought up that `text-width` really means `max-width` when the figure size is large (In reality it seems there's no major difference for captionWidth when the image size is large).

Here's my thought - what if we stick with this change - `text-width`. But then disable the caption width select in the UI if the size is Large, and make a mention in the docs that this property has no effect if size is Large.

zachberry:
Actually, it doesn't have any effect for size = medium either, since medium images already are at text width. So perhaps we disable the input for large and medium sizes.

jpeterson976:
Works great in the editor with my testing. Code looks good too, found some stuff I missed in the original PR. Only change I can think of is the new tutorial section should probably be updated to mention the caption width option is only available for small and custom sizes.

PR #1776, opened 2021-04-12
Increases cache time for module icons

iturgeon:
Despite previous efforts to implement eTags for cache, the browser still needed to ask the server if there was a new version every time.  The cache life time header was not set, so this does.  It'll keep the browser from even asking the server about the icon, which should drastically reduce requests


maufcost (approved):
Everything looks good on my end. Dashboard and modules render safely. Good job, Ian!

zachberry (approved):
LGTM!

PR #1777, opened 2021-04-12
Multiple page loading improvements

iturgeon:
* font `display: swap` avoids the browser displaying no text while font tags are loading.
* Uses `rel=preload` option to allow browser to load fonts and can prevent font repainting.
* editor inherited the viewer's initial loading animation, partial, and css
* script tags at bottom of pages moved to top and now using defer (can speed up time to render)
* inline js options removed from templates as they were not used

### DIFF
--- a/packages/app/obojobo-document-engine/setupTests.js
+++ b/packages/app/obojobo-document-engine/setupTests.js
@@ -2,7 +2,7 @@ require('obojobo-lib-utils/test-setup-chunks') // setup enzyme
 
 // Hack to get LaTeX to not warn about quirks mode:
 document.write(
-	'<!DOCTYPE html><body><div id="viewer-app"></div><div id="viewer-app-loading"></div></body>'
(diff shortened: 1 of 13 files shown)


maufcost (approved):
Everything looks good to me. Fonts and scripts loaded fine here. `defer` really helps DOM loading times as I've seen countless of times. 

One thing that I saw though is that on the new loading animation in the viewer, both the "Loading..." string AND the new loading animation appear when the module is loading (you can see that below). This doesn't prevent pages being loaded or loading times, so I'm LGTMing this :)

obo

zachberry (approved):
LGTM. I did a little testing on speed and measuring the total request time on a simulated Fast 3G network via chrome, and I did see some speed improvements

dev/19:
Editor: 8.48s
Viewer: 6.6s

This branch:
Editor: 8.35s 
Viewer: 6.3s

zachberry:
Numeric branch removes the "Loading...", so that branch plus this branch means you only see the one loading screen

Issue #1783, opened 2021-04-20
Clean up focus hacks with numeric questions, fix refocusing in the visual editor


zachberry:
The visual editor has an aggressive takeover of focus which you can see here - https://github.com/ucfopen/Obojobo/blob/master/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/visual-editor.js#L304

This is the reason in the numeric question inputs I have to do setTimeout hacks to get the focus restored. It potentially also is the cause of #1734.

I think we could come up with a less aggressive refocusing strategy that might solve these issues.

PR #1784, opened 2021-04-20
Issue/1783 clean up vis editor focus

### LINKED ISSUE #1734: Focus jumps around after using the Question Bank settings controls
These controls:

image

Steps to reproduce:
1. Create a tall page with a Question Bank inside of it somewhere
2. Click in some text far away from the QB
3. Click in one of the inputs in the QB settings
4. Click in some text close to the QB

Result: The page will jump to the previous text location in step 2

It has to do with `freezeEditor` and `unfreezeEditor`. In playing around, I was able to get good results with these changes:

```js
freezeEditor() {
	const el = document.activeElement
	setTimeout(() => {
		el.focus()
	})
}
```

Then `unfreezeEditor` isn't needed. However I haven't thoroughly tested it.

zachberry:
Relies on #1408 - That will need to be merged in first to review the code in the PR

Fixes #1734 and #1783

Changes:
The visual editor no longer refocuses itself onChange. That solves issues with focus problems with non-slate controlled inputs inside the slate editor. One fix was needed - The insert menu now manually refocuses the editor after its used, otherwise after using the insert menu the focus would be lost.


zachberry:
I added input validation here, so this also fixes #1792

maufcost (approved):
Numeric questions validation now looks really good to me. Any of the previous cases (including what @jpeterson976 and I found from #1408) seems to not break or show error messages anymore. I like that incomplete scientific notation expressions and unfinished fractions are being blocked now even before they go to the viewer.

Also, I'm still at odds with Slate, but the focus issue seems to be resolved now. I created a bunch of different types of nodes, and focus seems to get back to where it should be depending on the executed action. 

+ No lag issues on my end at least (differently from our most recent bug safari).

Great job Zac!

jpeterson976 (approved):
Focus issue looks to be resolved, and any error-producing strings were detected by the validation. I like how it doesn't even let you exit the input box if you have an invalid answer. It's still possible to save the document when it's invalid, though I have no idea how difficult/reasonable it would be to get around that, so LGTM!

PR #1787, opened 2021-04-20
Define order of items in the insert menu, clean up look

### LINKED ISSUE #1235: Hardcode order of items in the insert menu
The order of the insert menu is alphabetical but that doesn't lead to a good UX since common items like Text are far to the right, while lesser used items like ActionButton come first:

image

I think we should define the order of these things to put common items in front.

zachberry:
Fixes #1235 

This uses a config file in the obojobo-document-engine package - Not sure if this is the best solution, or correct place for it. I did try having it in the server package, which is where all of the other configs are, but it's not really a server config. I'm open to ideas.

Also lightens the look of (...) menus and makes it so the Materia icon turns purple on hover in the insert menu


iturgeon:
Configs in this case are for front end use, just in the editor - so it's a little different.  I don't think this is future proof, but it's a step in the direction. Does it break if Materia is disabled?

zachberry:
> Configs in this case are for front end use, just in the editor - so it's a little different. I don't think this is future proof, but it's a step in the direction. Does it break if Materia is disabled?

Good call, it likely does, but should be an easy fix

maufcost (approved):
LGTM! Great job, Zac!

PR #1799, opened 2021-05-03
Issue/89 red alert status

sergioncp:
red alert has been implemented with 100% test coverage


zachberry (approved):
Looks and works great! The only thing - your down migration needs to also remove the db indices it creates. Here's an example from another migration (this one just removes the indices, yours would remove the indices, then drop the table:

```js
exports.down = function(db) {
	return db
		.removeIndex('visits', 'visits_user_id_index')
		.then(() => db.removeIndex('visits', 'visits_draft_id_index'))
```

Since this is a training issue however I just want to point it out, no need to commit a fix.

zachberry:
Closing since this is a training issue

PR #1800, opened 2021-05-04
Created rubric summary

### LINKED ISSUE #1791: Create rubric modal summary
For reference: #1780

maufcost:
Fixes #1791


maufcost:
Additional idea: I also think that, because we're using those "magic strings" multiple times now in multiple files ("highest", "$attempt_score", "no-score", etc), we should make them constants in a different file to avoid errors, confusion, and duplicated code like we have in IframeSizingTypes and others.

zachberry (changes requested):
I really dig it, and I like your suggestion, I spawned it into an issue.

Two issues - I think the (?) gets a little lost, the gray background against the assessment rubric box background. I don't have any super great ideas, but maybe putting a slight border on the (?) would help - this is putting $color-action-minor around the button:

image

Second, I think I'd prefer the blue text in the rubric to start each sentence capitalized, even though that's grammatically incorrect I think visually it flows better.

zachberry (approved):
Awesome!

PR #1801, opened 2021-05-06
Fixed styled Materia caption

### LINKED ISSUE #1729: Materia caption gets mangled if it's styled and you open up the widget settings dialog
See:

maufcost:
The caption now keeps its style when you open and close a Materia chunk. Things to notice though that I thought you want to add, but currently is not in this PR:
- When the Materia settings dialog is open, its caption appears with all of its contents, but no styles are applied inside the dialog (it only shows the caption as a default text). The styled version of the caption appears as expected on the chunk in the editor outside the dialog.
- Authors can only edit the caption in the chunk and not inside its settings dialog. If you want me to implement something like a contentEditable for the caption inside the chunk's settings dialog, I'd be glad to.

Fixed #1729


packages/obonode/obojobo-chunks-materia/editor-component.js
```diff
@@ -68,13 +74,13 @@ class MateriaEditor extends React.Component {
...
 	// used wen returning from the settings dialog to set the title/caption
 	// using slate text so you can click on the caption to edit it
-	setChildText(text) {
-		const path = ReactEditor.findPath(this.props.editor, this.props.element.children[0])
-		// clear out text contents
-		Transforms.delete(this.props.editor, { at: path, unit: 'line' })
-		// set text into slate children text
-		Transforms.insertText(this.props.editor, text, { at: path })
-	}
+	// setChildText(text) {
```
zachberry, line 77: There is an issue with this, and it's hard to debug if you don't have Materia running on your system to test. But when you go to select a widget it will not only make the title of that widget the caption text in the dialog, but also makes that title the caption when you close the dialog. This change still shows the caption text as the widget title in the dialog, but when you close the dialog the actual slate caption underneath the node is blank.

zachberry (changes requested):
Good progress! But found an issue -

packages/obonode/obojobo-chunks-materia/editor-component.js
maufcost: Got it. That makes sense. Should I leave this issue for those who have Materia running locally then or is it possible to keep working on it without Materia?

PR #1803, opened 2021-05-06
List spacing branch

zachberry:


package.json
```diff
@@ -22,6 +22,7 @@
 		"test:ci:each": "lerna run test:ci"
 	},
 	"devDependencies": {
+		"babel-upgrade": "^1.0.1",
```
zachberry, line 25: What's this package for?

packages/app/obojobo-document-engine/package.json
```diff
@@ -61,9 +61,11 @@
 		"smoothscroll-polyfill": "^0.4.4",
 		"sysend": "^1.3.3",
 		"underscore": "1.10.2",
-		"url-parse": "^1.4.7 "
+		"url-parse": "^1.4.7 ",
+		"uuid": "^8.3.2"
```
zachberry, line 65: You should be able to use our internal uuid library instead, `import uuid from 'obojobo-document-engine/src/scripts/common/util/uuid'`

packages/app/obojobo-document-engine/package.json
```diff
@@ -61,9 +61,11 @@
...
 	},
 	"devDependencies": {
+		"@babel/plugin-proposal-class-properties": "^7.12.13",
```
zachberry, line 68: Needed?

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.js
```diff
@@ -6,6 +6,7 @@ import React from 'react'
 
 import MoreInfoIcon from '../../assets/more-info-icon'
 import TriggerListModal from '../triggers/trigger-list-modal'
+import ObjectiveListModal from '../objectives/objective-list-modal'
```
zachberry, line 9: Looks like the Objective stuff got mixed in with this branch, or this was based off of the Objective branch. Is it possible to remove the Objective stuff from this branch without too much pain?

packages/obonode/obojobo-chunks-list/editor-component.js
```diff
@@ -54,10 +54,11 @@ const List = props => {
 			/>
 		)
 	}
+	console.log(props);
```
zachberry, line 57:
```suggestion
```

packages/obonode/obojobo-chunks-list/editor-component.js
```diff
@@ -54,10 +54,11 @@ const List = props => {
...
 
 	return (
 		<Node {...props}>
-			<div className={'text-chunk obojobo-draft--chunks--list pad'}>
+			<div className={`text-chunk obojobo-draft--chunks--list pad li-spacing-${props.element.content.spacing}`}>
```
zachberry, line 61: Naming convention we're using for classes that are added dynamically via the react component are prefix them with `is-`, so I'm thinking instead of `li-spacing-*` we want `is-spacing-*` (i.e. `is-spacing-moderate`, etc)

packages/obonode/obojobo-chunks-list/editor-component.scss
```diff
@@ -1,6 +1,19 @@
 @import '~styles/includes';
 
 .obojobo-draft--chunks--list {
+	
+	.li-spacing-moderate {
```
zachberry, line 5: Better if these were more specific, so `&.li-spacing-moderate` - That ensures that the rule only applies on the root `obojobo-draft--chunks--list` element. Same comment for the other style.

packages/obonode/obojobo-chunks-list/viewer-component.scss
```diff
@@ -17,3 +17,15 @@
...
+	li {
+		margin: 0.5em 0;
+	}
+}
+
+.li-spacing-generous {
+	li {
+		margin: 1em 0;
+	}
+}
```
zachberry, line 31: Same with the editor-component.scss, these should be more specific - they need to be inside the root .obojobo-draft--chunks--list style as well. When all of the CSS gets compiled together by webpack the result is that all of the individual scss files are smashed together into one big file. Since these are by themselves not defined in any rule it means they'll impact anything on the page with the class of "li-spacing-generous". Unlikely to happen since it's a unique name, but want to avoid any chance of styles doing unexpected things.

zachberry: Additionally, reviewing the display of each of the list options in the viewer, they all look good except the moderate and generous options push the surrounding content around the list farther away. Adding some first-child and last-child rules to kill the top and bottom margins should fix it.

Compact:

image

Generous, note the two text lines above and below the list are father away.

image

FrenjaminBanklin:
This was taken care of in #1881.

PR #1812, opened 2021-05-10
Adjusts larger caption size for custom size figures to be text width

### LINKED ISSUE #1798: Larger caption size for images with size custom is actually at the max width instead of the text width
For small, med and large images if you set the caption width to extend past the image the text will start to wrap once it reaches the full content area width. However for custom images it wraps once it reaches the full content area (same width as questions). I didn't notice that in testing - we'll just need to change it so that for custom sized images it wraps like it does for the other sizes.

jpeterson976:
Fixes #1798

### DIFF
--- a/packages/obonode/obojobo-chunks-figure/viewer-component.js
+++ b/packages/obonode/obojobo-chunks-figure/viewer-component.js
@@ -14,7 +14,6 @@ const Figure = props => {
 	const content = props.model.modelState
 	const customStyle = {}
 	const captionStyle = {}
-	// console.log(content.captionWidth)
 	if (content.size === 'custom') {
 		if (content.width) {
 			customStyle.width = content.width + 'px'
--- a/packages/obonode/obojobo-chunks-figure/viewer-component.scss
+++ b/packages/obonode/obojobo-chunks-figure/viewer-component.scss
@@ -61,6 +61,8 @@
 			}
 
 			figcaption {
+				padding-left: 3.8em;
+				padding-right: 3.8em;
 				margin: auto;
 			}
 		}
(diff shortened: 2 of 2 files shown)


maufcost (approved):
This looks pretty good to me. There's one very tiny detail though when the caption extends past the figure's width, the caption is not exactly at text width:

figure
But this does not hinder any functionalities and still does what it is supposed to. Good job, @jpeterson976 !

zachberry (changes requested):
There's an issue with this at mobile phone sizes. The padding on text items changes at small widths. We have a `pad` class that sets padding values and those change depending on the browser width - I'm thinking instead of hardcoding these values here you'll want to conditionally give `figcaption` the `pad` class in the case where you have a custom image with a full-width caption

zachberry:
@jpeterson976 Actually, instead of adding the pad class, you can use the mixin - `@include pad();` - You can see it being used in the `&.small` rule.

maufcost:
With regular laptop screens, it looks pretty good to me. When I switch to mobile, I get this: 
Screen Shot 2021-05-12 at 9 30 46 AM
Is that difference between the text chunk's width and the caption on mobile intended?

zachberry:
That's my fault - I was wrong about `pad()`, it doesn't automatically handle mobile sizes like I thought.

I think this is what you'll want:

```scss
@include pad(1 / 0.8);

@include mq($until: desktop) {
	@include pad(0.75 / 0.8);
}

@include mq($until: tablet) {
	@include pad(0.3 / 0.8);
}
```

That mirrors the rules for `&.medium`, although compensates for the smaller font size in a caption (0.8em vs 1em)

zachberry (dismissed):
I noticed the .small size wasn't handling responsive padding either (this has been an issue for a while) so while I was here I added it. I also merged in dev/22. Anyway, looks good!

maufcost (approved):
Looking really nice now with these last changes! The padding on mobile is spot on and the caption behaves as expected with any size. Excellent job *claps*

PR #1815, opened 2021-05-11
Issue/1763 colored text not applying to list elements

sergioncp:
Text now picks up the color from the first colored element it sees


packages/obonode/obojobo-chunks-list/viewer-component.js
```diff
@@ -27,6 +27,26 @@ const addItemToList = (ul, li, lis) => {
...
 
+const pickColorFromText = (node, indent) => {
+	let bulletColor
+	let textStyles
+
+	try {
+		// needs try/catch because some nodes don't have a text element
+		if (indent === 1) {
+			textStyles = node.children[0].text.styleList.styles
+			bulletColor = textStyles[textStyles.length - 1].data.text
```
zachberry, line 38: `node.children[0].text` is an instance of StyleableText, and it has some internals you can use to grab the styles on the text there. Grabbing the last item from `textStyles` is dangerous because there's no guarantee that the last item is color.

Instead, you can do this - `node.children[0].text.styleList.getStyleComparisonsForRange(0, 1, StyleType.COLOR)` - That will return an object that details where that style is, like this:

```js
{
    "after": [],
    "before": [],
    "enscapsulatedBy": [],
    "contains": [
        {
            "type": "color",
            "data": {
                "text": "#990000"
            },
            "start": 0,
            "end": 3
        }
    ],
    "left": [],
    "right": []
}
```

You'll want to look at `contains`. That means that the text from index 0 to 1 contains the color style. In which case, you can dig into `contains[0].data.text` to get the color.

packages/obonode/obojobo-chunks-list/viewer-component.js
```diff
@@ -42,8 +62,14 @@ const renderEl = (props, node, index, indent) => {
 			)
 		case 'element': {
 			const ElType = node.type
+			const bulletColor = pickColorFromText(node, indent)
```
zachberry, line 65: `node` is an instance of MockElement, and as you figured out it's a nested structure. There's another data structure you can query that will be easier - `props.model.modelState.textGroup` is an instance of `TextGroup`, and it will have a `TextGroupItem` for each item in the list. That data structure isn't nested - if there are four items in the list then you'll have four `TextGroupItem`s and you can get each one by doing `props.model.modelState.textGroup.get(index)`. To get the StyleableText you can do `props.model.modelState.textGroup.get(index).text`. That you could pass into `pickColorFromText`, and `pickColorFromText` just needs to dig into the style of the first character as I commented above.

zachberry (changes requested):
Great so far, I have some suggestions that use some of the more internal not-well-documented features that should handle some of the manual work that you had to do in the viewer code.

packages/obonode/obojobo-chunks-list/viewer-component.js
sergioncp: That will definitely make things easier, and safer, I'll use textGroup and might have to change the function a bit but the above comment will still help

sergioncp:
After some testing I noticed that a MockElement of type "li" first child, `node.firstChild.text` seems to always return the same StyleableText as `props.model.modelState.textGroup.get(index).text`. The function `renderEl` will always run more times than the total number of StyleableTexts inside of `props.model.modelState.textGroup`, say 7 times for a 4 indent list, making it a bit difficult to track the proper index of the text that needs to be accesses without heavily modifying the `renderEl` function since both the `index` and `indent` passed seem to be tracking relative to something else

PR #1821, opened 2021-05-13
Fix duplicate nested score action pages being created

### LINKED ISSUE #1689: Clicking enter on the button on the Score Actions page duplicates the score action page
Shown here:

zachberry:
Fixes #1689 

The issue: If you make a new empty module, go to the score action page and then focus in somewhere in the Retry Assessment button, then hit enter, a nested duplicate score action page is created.

The reason: Buttons use the breakToText method in KeyboardUtil. Normally hitting enter in a node causes slate to duplicate and split that node. In this case you'd get two Buttons. But this is overridden with breakToText, which turns the second sibling node into a regular Text Chunk node. The issue is that the newly created Text Chunk node is given children which, normally, is just a text node like `{ text:'abc' }`, but in the score action page is actually a whole Score Action node.

The solution: Find all of the text nodes of the split sibling child and put those as the children of the newly created Text Chunk. In the case of the score action page, that means it will find the leaf `{ text:'abc' }` type nodes and use those rather than nest a whole Score Actions node.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/util/keydown-util.test.js
@@ -365,7 +365,11 @@ describe('KeyDown Util', () => {
 			children: [
 				{
 					type: HEADING_NODE,
-					children: [{ text: 'someText' }]
+					children: [
+						{ text: 'someText' },
+						{ children: [{ text: 'and' }] },
+						{ text: 'someMoreText' }
+					]
 				}
 			],
 			selection: {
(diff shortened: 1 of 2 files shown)


maufcost (approved):
So nice to see how this issue was resolved (I spent a couple hours trying to solve it, but to no avail).

Everything looks pretty good to me. I tested it by selecting different parts of the text in the button and hitting enter, different cursor positions, changing button ids, and checking if actions would be carried to the viewer.

Awesome job. LGTM!

jpeterson976 (approved):
LGTM! Couldn't get the page to duplicate itself any way I could before.

PR #1822, opened 2021-05-13
Fixes convert-if-list to work in any nested structure

### LINKED ISSUE #1819: Automatic list creation by typing "* " not working in Questions
I see an error, something about children not being defined

zachberry:
Fixes #1819 

The automatic list creation when typing "*" or "2. " (for example) was breaking inside Questions. The `nodeIdx` variable was hardcoded to look for the second item in the path. I changed this to look at the second from last item in the path, which should make it work in any nested node structure.


jpeterson976 (approved):
LGTM! In my testing I noticed that if you add a list in the Score Actions, the text will be centered instead of left-aligned with the bullet, which I figured could be its own separate issue since it's unrelated to this one.

center-aligned list text

PR #1824, opened 2021-05-17
Multi-select modules in the Dashboard

### LINKED ISSUE #1790: Allow multiple modules to be selected in the Dashboard for bulk operations
It can be a pain to (for example) share several modules or delete several modules once at a time. It would be nice to be able to select multiple modules to perform bulk operations.

jpeterson976:
Resolves #1790 

No longer WIP - lets users select and delete multiple modules at once. Sharing/downloading will be added on with future issues.

Todos:
- [x] Dashboard UI for multi-select mode
- [x] Modify redux actions for selecting/deselecting multiple modules
- [x] Add new action for bulk deletions
- [x] Have UI actually do things
- [x] Jest stuff


maufcost:
That is a great start, @jpeterson976 :) The design looks pretty good to me. There is just one thing I'd like to say though: I believe leaving those checkboxes with their full opacity at all the times kind of clutters the dashboard. We could make them appear on hover. I tweaked the code a bit and got this:

obo

zachberry (changes requested):
This is super awesome!

I haven't reviewed the code yet, but here changes I'm thinking:

1. Allow users to hit ESC to cancel out multi-select mode
2. I'm thinking initially only show the checkbox when the module edit menu is open rather than on hover
3. Once multi-select is enabled, no longer show the module edit menu 
4. I'm wondering if we need a more annoying confirmation when deleting modules - as in, they are asked to type DELETE to confirm they want to delete multiple modules at once. If we don't, then I think the warning prompt should be more serious, like "Are you sure you want to DELETE these N modules? This action can't be undone!"
5. You replicated the mock up perfectly, but seeing it in action I'm worried the Delete and (X) buttons are too hard to notice. Here's a mockup of what I'm thinking - I don't like the look of this as much but the change to this purple background which matches the multi-select background of selected modules is a more noticeable change:

image

zachberry (approved):
Nailed it. LGTM!

maufcost (approved):
This looks so good! Can't wait to see it in production. Tested by deleting a couple modules all at once and typing variations of "DELETE" (e.g. "Delete", "delete", etc). Everything works as expected, and as always, great job!

paradisojr:
What bulk actions will be available with this feature? Will users be able to bulk download or clone? I need to clone 50+ modules for a faculty member this term and that option would be really helpful.

zachberry:
> What bulk actions will be available with this feature? Will users be able to bulk download or clone? I need to clone 50+ modules for a faculty member this term and that option would be really helpful. I suppose a bulk share would be vital, too.

Not yet, but it's coming. This feature will let you bulk delete, and then we'll add in other bulk features in another update.

PR #1828, opened 2021-05-20
Updated iframe properties dialog

### LINKED ISSUE #1536: Updates to IFrame properties visual editor dialog
What we have now in the editor to edit/set iframes was a first step so some of these aren't surprises. That said, the biggest pain for authors adding in iframes is:
1. Don't know if it's setup correctly / working until you preview
2. Lots of options, some which aren't clear
3. In the XML/JSON you can specify the type of content, which sets defaults for how to display the iframe - there's "media" and "webpage". Those don't play super great with the visual editor UI
4. If you want to embed (for example) vimeo you'd need to get the <iframe embed code, then know enough HTML to pull out the "src" attribute, and then copy the width and height into our UI.

Suggesting a revamp of the dialog to solve these problems:

Editing a brand new iframe:

image

Here we let you paste the iframe embed code from a site, and we'll auto extract all the relevant information - really just "src", "width" and "height". There's also an assumption we'll make here - if you paste in <iframe> code we'll assume you're probably embedding media, and if it's just a regular URL you're probably embedding a webpage. Once you paste in something this first dialog could expand:

image

Here we go ahead and try to embed what you pasted just so you can confirm it's working. If they click the "No" link here it could expand to explain that some pages may restrict their content inside an iframe, or, to make sure if you're trying to embed media to use the iframe code and not the regular url

If you click Yes you'd get to this step:

image

> (I didn't draw having one of the Content type options selected, but it one of them would be selected in the actual UI)

Here if you click 'Change' it would take you back to the previous page. We'd also go ahead and pre-select Content type for you, based on the assumption I mentioned from earlier.

Assuming that 'Video or other media' is selected, the rest of the dialog is a smaller subset of options. Sizing would define the dimensions of the iframe. The first choice - full width - sets the iframe to the maximum width (same width as questions, which extend past the text). The second choice is similar, it just maxes out at the same width as the text. Finally the last option uses the exact width.

IF the iframe that they copied had width and height values then we can determine the ratio of the iframe - typically for video that will be around 16:9. If you pick the first two responsive options we basically set the width to 100% and use the ratio to determine the height (the iframe node already does this stuff) - but then we don't really need to show you the Width and Height inputs since they are just needed for the ratio. However the 'Edit dimensions' link would let you change those values.

If you picked the Fixed size option you'd be shown the width and height inputs.

If you paste in <iframe> embed code but there isn't a width and height then we can't determine the ratio. In that case, we'll assume a 16:9 default with a size of 600 * 338.

Next option - Loading - would replace "autoload". Using a select box instead of the toggle just so that we can explain the different choices, and in the example you can see a warning about why the auto option might be negative. Additionally, we could add in a lazy-load option, if we use IntersectionObservers to determine when the iframe is on the screen.

IF you pick one of the auto-load options you don't need to set the title, so when selecting one of those options we'd hide the title input.

FINALLY, you can click "Show all options" which would give you every option that we're hiding (such as which controls to display, etc). I don't have that UI mocked up here.

IF you paste in an YouTube link we'd notify you that we have a special display for YouTube video and ask if you want to convert this to use the YouTube chunk instead. That UI isn't mocked up (This could be moved to a separate issue)

Second, if you paste in a URL (not <iframe> code) and we assume you're embedding a webpage the sizing options are a little different. In that case the width is always fixed, and the height is what can change. So instead of "Sizing" we'd probably switch that label to "Width", since it describes the width. below that we'd have an input for height, since we do need that - but the placeholder text would be set to the default if you don't set anything (which I believe is 500px?). If you picked the Fixed size option we would then add an additional input for you to set the width in px.

maufcost:
Fixes #1536


maufcost:
A sneak peak at our new iframe properties modal:

iframe

jpeterson976 (changes requested):
This looks great! I especially love the sizing options, those SVGs look awesome. Great work!

One thing is that in the advanced options, the description text can be overlapped by the toggle button if the window isn't wide enough (shown below). 

Screen Shot 2021-06-09 at 3 18 18 PM

Also, this may just be a me thing, but I think that the labels in the 'Options' section should be vertically centered with respect to their inputs as well. 

This is probably pretty nitpicky, but I think the 'Content type' label should be aligned with the options box - currently it's aligned with the container that contains the description text so it's sitting slightly below where it seems like it should. Similarly I think the 'Sizing' label should be aligned with the middle bullet point, though getting them to properly line up there might be annoying. (If anyone disagrees with this, I'm open to being convinced otherwise)

zachberry (changes requested):
Super excited about this, sorry for the delay in the review. I think you have the base of the form working pretty well, but I do have a bit of feedback on changes:

First screen:

* The preview input needs a debounce. If I try to type in a URL as soon as I input one character the iframe attempts to load right away (1)
* If you give the input an invalid URL, it will instead load the editor inside the iframe. Likewise, checking if the input is an iframe embed code could be simplified by trying to put the users' text as innerHTML inside a div and poking into the div to see if the browser created a valid iframe. You can see an example of both of these techniques inside the "parse-youtube-url.js" file, in the `parseYouTubeURL` function. It might be worthwhile to extract this code and generalize it to be able to be used in both places. (2)
* Add a 1px $color-shadow border to the iframe preview (3)
* The "Does the preview look good?" help-text state expands the dialog to suddenly be very wide. Want to make sure the width doesn't change between these states. (4)

Second screen:

* The Content type and Sizing boxes need `cursor:pointer` and a hover state. (5)
* The "px" in the Fixed size sizing option as well as the "!" in the warning icon under the Loading select box is using some default-level font. If these are in the SVG you'll want to use Libre Franklin (free to download) and then export the SVG with the fonts rendered as shapes. (6)
* Dialog needs a min-width so that the helper text below the Content type boxes doesn't wrap (7)
* Switching the different Content Type choices should set the options below automatically. By default 'Video or other media' would set Fit to Scale, Reload turned on and the other controls turned off. Embedded webpage should set Fit to Scroll and enable all controls. Additionally Sizing should default to Responsive - text width. (8)
* The text "Load when student sees IFrame" should be changed to "Load when student clicks IFrame" (9)

General:

* Buttons and inputs seem to be using Arial - want to use the standard Selects, Buttons and inputs for these. The Materia dialog uses SettingsDialogForm, that could be of use - It helps build out forms - Ian added it in the Materia branch. (10)
* When displayed in the viewer there seems to be a gray background to iframes - If a page doesn't define a background color it's more noticeable. (11)
* Editing an iframe should jump to the second screen if the iframe has a URL set (12)

maufcost:
I numbered the issues on Zac's last comment so they are easier to address and talk about.

**(1)** In order to add a debounce with our current library (`use-debounce`), we would need to transform our `EditIFrameModal` from a class-based component to a function component in order to work with the debounce hook. Is that ok if the `NewIFrameModal` is still a class-based component and the `EditIFrameModal` is a function one? Or do you prefer me to transform both of them to function components?
**(2)** I relied on the `parseYouTubeURL` to check for valid iframe embed codes and urls and everything works fine now, but the function turned out to be a lot simpler than the one for the youtube node. So, instead of joining them together, I was thinking about moving that file up some lex levels, and make both the iframe and youtube chunks import from the same file. Any thoughts?
**(3)** Done.
**(4)** Done.
**(5)** Done. I just need some feedback on the purple borders. Any thoughts about keeping the purple borders all the time or only on hover?
**(6)** Done.
**(7)** Done.
**(8)** Done.
**(9)** Done.
**(10)** Done. Applied the mixins and adapted inputs and selects to have the same border color.
**(11)** I haven't found any background colors set on iframes in the viewer. Could you send me any screenshots? :)
**(12)** Done.
**(13)** Proposing this one: I thought about disabling the button to go from the first modal to the second if the `src` typed is not valid. Any thoughts?

maufcost:
Finished applying the changes Zac requested. This PR is testable/reviewable again.

The tests that are not passing are some uncovered React functions I had to write in order to control more complicated hover/click actions. I'm just waiting to get your feedback on the current hover/click states I implemented before writing them.

jpeterson976:
Looks good so far! Only thing I noticed is that setting it to "Embedded webpage" leaves out the Border setting when setting everything to true. Everything else seems to be working as expected with my testing. Good work!

For your proposal with (13), I think that it could be useful since it'd help prevent accidentally giving an invalid src and not realizing it. I don't think there are any cases where an invalid src would be intentional, but if they do exist then an extra warning message similar to the missing src message might be a better option.

Also, #1875 was recently merged in and removes the leading comma from the `controls` strings, so those should be safe to remove in the code now.

maufcost:
Thanks for all the suggestions so far everyone!

New changes include:
* Addressed the last changes that @jpeterson976  requested.
* I also found a handful of bugs with how `content types` and `controls` were handled. For that, I merged the `border` property into the `controls` property, allowing me to refactor a lot of code and better handle how the entire `controls` state is managed.
* I introduced a new property to the iframe context, called `controlsChanged` to better address Zac's **n#8** issue above. I was facing a bit of a boolean dilemma setting the controls to have specific properties depending on their iframe's content type _(because we would also need to know when users created iframes, turned off all controls, and closed them; created iframes, left the controls with their default options, and closed them; and all kinds of combinations of these user actions)_. This new prop solved this issue (apparently).

Let me know your thoughts :)

jpeterson976 (approved):
LGTM! Did some playing around with changing the settings, everything looks to be working as expected. Great work!

FrenjaminBanklin (dismissed):
After merging in changes from dev/31 and updating snapshots, this looks like it's running well.

FrenjaminBanklin (approved):
Merged in the current dev branch and checked functionality, it looks like all of this is working really well.

I did also contribute to this so it should probably have another person besides me look it over before I finalize it.

RachelDau:
These iframe options are really cool. I didn't find any issues with anything. I tested on different browsers and played around with the sizing and other settings.

RachelDau (approved):

PR #1832, opened 2021-05-25
Constants for Assessment Rubric

### LINKED ISSUE #1802: Use constants for Assessment Rubric "magic values"
Based on a comment from @maufcost in #1800 :

```
...("highest", "$attempt_score", "no-score", etc), we should make them constants in a different file to avoid errors, confusion, and duplicated code like we have in IframeSizingTypes and others. ...
```

jpeterson976:
Fixes #1802 

I noticed that `assessment-rubric.js` already has most of the needed constants defined within it, so I added the remaining ones and modified the rubric modal to use them where applicable.


maufcost (dismissed):
Now that's looking good. Excellent!

I opened the modal with different settings just to check if there were any typos or something that would somehow break it, but everything is good :)

zachberry (changes requested):
I see a few other cases where these could be used. In the assessment obonode package there's rubric/converter.js and editor-component.js, but also in the viewer package there's get-display-type.js (inside assessment-score-reporter).

Here's what I'm thinking (please don't kill me):
Option 1: Move assessment-score-reporter into the assessment obonode package (which it should be anyway, unless I'm missing something). Then, add the constants to rubric/converter, editor-component and get-display-type.
Option 2: Just add the constants to rubric/converter and editor-component, and make a separate issue to move assessment-score-reporter and add the constants.

zachberry (approved):
Tried a lot of different rubrics, seemed to work great in all cases!

PR #1842, opened 2021-05-27
youtube events added

sergioncp:
youtube events have been added, `seekTo` event relies on `played` events


zachberry (changes requested):
First off, this is awesome! I've been wanting this feature for a loooong time.

I haven't reviewed the code yet but I do have some changes:

1. These are nitpicks, but I'm thinking change `playHead` to `playhead`
2. Nitpick: Standardize the tense of the event names. `media:played` => `media:play`, and `media:buffering` => `media:buffer`
3. I think we could keep track of the watch time and include that in the event under `secondsWatched`. That would simplify the event processing needed for users reviewing the data.
4. We should add a URL, `url: "https://youtu.be/dQw4w9WgXcQ"`
5. We should add the obo node ID of the video chunk, perhaps as `nodeId`.
6. Need to capture when the video stops playing because the user changes the page. I'm thinking this could be `media:unload`.

As a summary of the event data changes above, here is a before and after example (this is a `media:play` event):

```js
const before = {actor: "user", playHeadPosition: 0}
```

```js
const after = {
  actor: "user",
  playheadPosition: 20,
  url: "https://youtu.be/dQw4w9WgXcQ",
  nodeId: "547BA3C2-BA55-419D-B5F8-9B6DA31D0044",
  secondsWatched: 15,
}
```

packages/obonode/obojobo-chunks-youtube/youtube-player.js
```diff
@@ -94,6 +131,146 @@ class YouTubePlayer extends React.Component {
 		)
 	}
 
+	onStateChange(playerState) {
+		const currentPlayheadPosition = Math.floor(this.player.getCurrentTime())
+		const previousSecondsWatched = this.currentState.secondsWatched
+
+		this.previousTime = this.currentTime
+		this.currentTime = Date.now() / 1000
+		const currentSecondsWatched = previousSecondsWatched + (this.currentTime - this.previousTime)
```
zachberry, line 140: I want to do this calculation in media store instead of here. That way you can send media store events like play, pause, etc, and it'll figure out how much time has been spent watching.

I also noticed a bug - When changing the page there is a media:unload event but the seconds watched still gets updated with unexpectedly big values

zachberry (changes requested):

zachberry (changes requested):
I attempted to get this pushed in for dev/25 and overall it's working good, but there still are some oddities that make me want to hold off for now. I think the previousPlayheadPosition value isn't quite aligning with what I was hoping it would be, so I'm thinking we ditch it.

Also, the seekTo event is working to the best of what @sergioncp could do with what little the YT API gives us, but in testing it in depth I think it's too kludgy. I think what I'd want is if a user skips ahead in the video the event stream would be

* User hits play -> media:play event is fired
* User skips to 20 seconds in -> media:seekTo event is fired
* Video is buffering -> media:buffer event is fired
* Video starts playing -> media:play event is fired

However in actuality a pause event is fired and the seekTo event replaces the play event. I think that makes it difficult to understand the data.

Lastly, I still felt there was some oddities with the actor property not always aligning with what I'd expect. So yet again, I'm thinking (at least for these 1.0.0 events), we ditch it.

I think what might work best in the end would be to simply send events as they come in from the YT API state changes. I really would like to smooth over the data but I think it's a bit too muddy and we can scale back for the 1.0.0 events.

PR #1843, opened 2021-06-01
Fixes input box weirdness for numeric questions w/ large units text

### LINKED ISSUE #1826: Numeric questions input box weird when units text is large
Likely not a common case, but we should have a max-height defined for the text input on numeric questions - See this example:

image

jpeterson976:
Resolves #1826

### DIFF
--- a/packages/obonode/obojobo-chunks-numeric-assessment/__snapshots__/viewer-component.test.js.snap
+++ b/packages/obonode/obojobo-chunks-numeric-assessment/__snapshots__/viewer-component.test.js.snap
@@ -86,6 +86,9 @@ exports[`NumericAssessment NumericAssessment renders a correct response 1`] = `
     <div
       className="input-container"
     >
+      <div
+        className="result-symbol"
+      />
       <input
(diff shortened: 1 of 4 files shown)


zachberry (changes requested):
One final change, would like the checkmark or X to also be centered, seen here:

image

zachberry (changes requested):
I found an issue when feedback is involved:

image

Otherwise, I tested various examples of both types of questions, both with and without feedback and with and without lots of text and all other cases look good.

jpeterson976:
I noticed that the review flags when you click "Reveal Answer" were also not aligned properly, so I fixed that as well. I added some adjustments to the more info button for smaller screen sizes too, since it was hidden by the input box at sub-tablet sizes.

zachberry (approved):
I tested this with numeric and MC questions, both in practice and assessment, at all mobile sizes, and tested with ChromeVox, and found no issues - looks great!

maufcost (commented):
Ignore this comment (Was supposed to be a review...) (Review below)

maufcost (approved):
LGTM! Tested with the same tools and sizes as Zac above ^ Good job!

PR #1844, opened 2021-06-01
Id input in menu disabled by default

### LINKED ISSUE #1701: `id` in (...) menu too easy to accidentally edit
The most common case of opening the (...) menu is to edit page titles - when you're moving quickly it's very easy to accidentally put your title in the id field (since it's the first item focused on when the dialog opens).

I'm thinking we hide the ID away in a click, so they have to click a link button to edit the ID. Since most authors won't want to mess with the id, it'll also help clean up the dialog since the id can look intimidating in it's normal UUID format.

jpeterson976:
Resolves #1701

Giving the input box the actual `disabled` tag makes it unable to get focus which caused a couple issues, so I made it `readOnly` and added some style rules to make it imitate the look of a disabled input.


zachberry (changes requested):
This is just a review of the front-end. I noticed a few things.

For me the buttons are getting pushed off to the next line (probably want to transition to flexbox):

image

* Not sure if it's possible, but if so I'd like to keep the details menu open after they confirm the dialog that pops up notifying them that the ID was copied (An alternative for the dialog, stolen from github, would be to change the button text from `Copy Id` to `Copied!` for a few seconds)

jpeterson976:
I fixed the buttons displaying incorrectly, turns out I moved around some CSS rules right before committing and didn't check if it affected how things looked.

The clipboard thing would require modifying `ClipboardUtil` since the modal created inside of the function, so I've just left it as it is for now since then a bunch of other things would have to be modified too.

zachberry (approved):
Ah, valid point. LGTM!

maufcost (approved):
This is looking pretty good to me! Great work :)

PR #1856, opened 2021-06-09
Dashboard group selecting behavior fixes

jpeterson976:
Based on #1824

The last selected index is now reset when the sort order changes and when leaving multi-select mode, which should resolve the weird group selecting behavior that we found.


zachberry (approved):
Tested, works well!

PR #1857, opened 2021-06-09
Limit iframe text length

### LINKED ISSUE #1837: Limit IFrame text
When not specifying a title and using a very long URL the result feels janky:

image

This will show up in both the editor and viewer although it's handled slightly better in the viewer. I think we just need to cut off the amount of text that is displayed (ideally using an ellipsis).

jpeterson976:
Fixes #1837 

I limited the length to 50 characters, which I chose somewhat arbitrarily, so let me know if another max length is preferred.

### DIFF
--- a/packages/obonode/obojobo-chunks-iframe/editor-component.js
+++ b/packages/obonode/obojobo-chunks-iframe/editor-component.js
@@ -71,8 +71,6 @@ class IFrame extends React.Component {
 	}
 
 	getTitle(src, title) {
-		const charLimit = 50
-
 		if (src === null) {
 			return 'IFrame missing src attribute'
 		} else if (title) {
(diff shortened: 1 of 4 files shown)


packages/obonode/obojobo-chunks-iframe/editor-component.js
```diff
@@ -71,13 +71,21 @@ class IFrame extends React.Component {
 	}
 
 	getTitle(src, title) {
+		const charLimit = 50
+
 		if (src === null) {
 			return 'IFrame missing src attribute'
 		} else if (title) {
 			return title
```
zachberry, line 79: This also needs to be limited - This code is ran if the user supplies their own title

zachberry (changes requested):

zachberry (approved):
LGTM!

PR #1864, opened 2021-06-21
Adds assessment stats for modules and new /stats page in the repository

zachberry:
**⚠️ Note: This branch has migrations and new packages! ⚠️**

* Added canViewSystemStats permission. By default this stat can't be added via an LTI launch, it must be applied manually
* Added user_perms table, which unilaterally applies a permission (like canViewSystemStats) to a user
* Added a new Assessment Stats dialog to modules in the repository. This allows you to see a score breakdown of students' final assessment scores and their attempt scores.
* Added /api/assessments/:draftId/details, a new API endpoint to power the new assessment stats page
* Added a new /stats page to the repository, which allows users with the canViewSytemStats permission to be able to view assessment details for any draft in the system
* Added /api/drafts-all, which returns draft summary information for all drafts in the system

Additional things:
* Adds the year to the version listing on the version history dialog
* Changed the name of AssessmentModel's fetchAttempts to fetchAttemptsForUserDraftAndResourceLinkId
* Changes User so that instead of methods like user.canPreviewDrafts which are created dynamically, now user has a perms array, and you would use user.hasPermission('canPreviewDrafts')
* Removes the 'Dashboard' link from the repository if the user isn't logged in

To test:
* Click on the More... button for a module in the repository, then click on the new Assessment Stats button
* Add a record in user_perms for the Test Instructor 1 user, with the permission `{ canViewSystemStats }` - This should add a new Stats option in the repository header nav bar


maufcost (changes requested):
Just found one issue:
1. Go to the dashboard and click on the Assessment Stats of a module (complete at least one assessment on that module just so that you can see at least one row on the table).
2. Then, on the bottom, include all the data and advanced fields before filtering or searching anything.
3. When selecting Search by `Course Title`, type something random (e.g. "asdasda") on the text input that opens up (No records will be displayed as expected).
4. Now, even if you delete that entire random string, the table will still show that there are no records to display. I believe the expected behavior here would be to display all rows when there's nothing typed on the search input.

Appears to be something with the debouncing:
gif1

Aside from that, the Assessment Stats component seems to be working wonders *chef's kiss*

The year on the version history dialog - Works!
Dashboard link removed when user is not logged in - Works!
The user_perms, canViewSystemStats, and the new Stats option appear as expected. Modules can be selected and their stats viewed.

Jeez. Amazing work, Zac!

PR #1866, opened 2021-06-22
Fixes strange rubric rendering with set passing score

### LINKED ISSUE #1863: If rubric is set so that it sets your passing score to a set value, it renders strangely
This is the result of a rubric that sets your score to 90% if you pass. The explanation doesn't make sense: 

image

Here is the exact rubric

```xml
<rubric type="pass-fail" failedResult="0" passedResult="90" passingAttemptScore="100">
	<mods>
		<mod reward="4" attemptCondition="[1,$last_attempt]"/>
	</mods>
</rubric>
```

A similar confusion can arise when no mods are present:

image

The readout for a score adjusted for failing makes more sense:

image

jpeterson976:
Fixes #1863 

It didn't seem to like when the attempt score was 100 but the rubric sets the score to >100 when passing, so I added a new display type for that case that gives the same info as when the attempt score is less than 100 but passes.


zachberry (changes requested):
There might be a simpler way to tackle this - I'm thinking we merge these:

```
TYPE_PASSFAIL_PASSED_GIVEN_SCORE_AND_ATTEMPT_SCORE_LESS_THAN_100,
TYPE_PASSFAIL_PASSED_GIVEN_SCORE_AND_ATTEMPT_SCORE_IS_100_AND_NO_MODS_REWARDED,
TYPE_PASSFAIL_PASSED_GIVEN_SCORE_AND_ATTEMPT_SCORE_IS_100_AND_MODS_REWARDED,
```

Into just one: `TYPE_PASSFAIL_PASSED_GIVEN_SCORE`

And it would replace the case in the switch for `TYPE_PASSFAIL_PASSED_GIVEN_SCORE_AND_ATTEMPT_SCORE_LESS_THAN_100`.

That way any time a score is strictly set for meeting the threshold score it would always render as

```
Attempt Score (Passed): xx%
--------
Score adjusted for passing: yy%
```

I think the code as it is is attempting to avoid situations where the set score is equal to the attempt score (So students don't see text like `Score: 100%, Score adjusted for passing: 100%`), but with the ability of hindsight I think that was probably a bad decision. If we don't care about it, it will simplify the code and remove this problem

zachberry (approved):
I tried all the combination of rubrics I could think of, all working great. I merged in the latest dev/25 changes and assuming the tests past I'll merge in

PR #1876, opened 2021-07-07
Page jumping when answering a question

### LINKED ISSUE #1808: Answering very tall questions can cause the page to jump around
In this example the question is pretty tall with a figure and the top portion is obscured. Once answered the page jolts so that you jump up to see only the top half of the question:

jpeterson976:
Resolves #1808

Swapped `preventScroll` to `true` when answering a question, so it should no longer jump around but still change focus properly.


maufcost (approved):
LGTM. Works both on regular modules and tutorials. Tested with single-answer and multiple-choice questions by creating unnecessarily gigantic questions. As far as my testing goes, the page is not jumping anymore to the top.

Good job, Jacob!

zachberry (approved):
Nice.

PR #1878, opened 2021-07-12
Prevents scrolling in repository when a modal is open

### LINKED ISSUE #1862: Prevent scrolling of body in repository when modals are displayed
A minor enhancement, but when using the mouse wheel when a modal is displayed in the repository the page in the background scrolls up and down, which is unexpected.  if we set body to `overflow: hidden;` the page won't scroll. Perhaps there's an easy way for us to programmatically set this when a modal is shown.

jpeterson976:
Resolves #1862


zachberry (approved):
Nice elegant solution

PR #1887, opened 2021-07-25
Set up dependency Cruiser 

iturgeon:
In an effort to simplify dependencies, simplify testing, and and minimize [over use of DRY principles](https://qvault.io/clean-code/dry-code/), it's valuable to visualize and build rules over dependencies between modules  These tools can be used to both visualize that and add tests that can be run automatically to enforce simplicity. 

adds
* test:deps - can be used in ci pipeline to block merging if dependency cruiser rules fail
* deps:high - build high level dependency visualization 
* deps:detail - build detailed dependency visualization

Requires `brew install graphviz` on macOs to get 'dot' 

This is just a small initial idea inspired by https://youtu.be/TqfbAXCCVwE?t=902

### DIFF
--- a/.dependency-cruiser.js
+++ b/.dependency-cruiser.js
@@ -1,0 +1,421 @@
+/** @type {import('dependency-cruiser').IConfiguration} */
+module.exports = {
+  forbidden: [
+    /* rules from the 'recommended' preset: */
+    {
+      name: 'no-circular',
+      severity: 'warn',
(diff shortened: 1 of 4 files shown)


iturgeon:
Some examples:

Screen Shot 2021-07-25 at 5 21 41 PM

Screen Shot 2021-07-25 at 5 22 52 PM

PR #1889, opened 2021-07-29
Added support for colored list indices and bullet points

### LINKED ISSUE #1763: Colored text not applying to list elements
See here how the 1., 2., etc parts of the list remain black:

image

maufcost:
I had to change my approach a couple times to facilitate testing.

Besides all the precedence CSS/Sass I dealt with, I ended up creating a new function on obo's `MockElement` class to give us the supposed color of list indices and bullet points (right now, in order to be more careful and as expected, this function only works with list elements with text inside since we may have other types of elements as list elements).

As always, I'd be glad to address any requested changes.

Fixes #1763


jpeterson976 (approved):
LGTM! Seems to work really well in my testing, I didn't come across any issues. Good work!

yarn.lock
```diff
@@ -14282,6 +14282,11 @@ url@^0.11.0:
     punycode "1.3.2"
     querystring "0.2.0"
 
+use-debounce@^7.0.0:
```
zachberry, line 14285: Is this needed?

zachberry (changes requested):
Works great! I just had one question about a package

yarn.lock
maufcost: That is weird. I didn't install it. Is it possible that somehow it installed it when I ran `yarn install` when I cloned the repo?!

zachberry: It's possible (and likely lol) that it might be a mistake on a previous PR I made, and neglected to commit the yarn.lock file. I'm okay merging it in then

maufcost: It's up to you. I can try downgrading it if you want me to.

zachberry (approved):

PR #1892, opened 2021-08-09
Backspace at start of list converts it to text

### LINKED ISSUE #1786: Hitting Backspace at the start of a list should convert it to a regular text chunk
Currently it simply deletes the node, which feels unexpected if you're used to Google Docs. It can also be frustrating if you type "* " and it automatically creates a list, then you want to undo that so you hit backspace, then the chunk is simply deleted.

jpeterson976:
Fixes #1786 

Hitting backspace at the start of the first list item should now convert it to a text node when at the highest level. It'll also make Slate keep the cursor at the start of the node after the conversion instead of at the end. I made a slight modification to `convert-if-list` as well to do the same as I noticed it would put the cursor at the end of the line, which seemed unexpected.

### DIFF
--- a/packages/obonode/obojobo-chunks-list/changes/on-backspace.js
+++ b/packages/obonode/obojobo-chunks-list/changes/on-backspace.js
@@ -2,6 +2,8 @@ import { Editor, Transforms, Range, Node } from 'slate'
 
 const LIST_LEVEL_NODE = 'ObojoboDraft.Chunks.List.Level'
 const LIST_LINE_NODE = 'ObojoboDraft.Chunks.List.Line'
+const TEXT_NODE = 'ObojoboDraft.Chunks.Text'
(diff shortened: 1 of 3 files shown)


maufcost (changes requested):
Look at you mastering Slate! Awesome work.

I found one edge case wherein if we convert a list from unordered to ordered and try to convert the list back to a text node by hitting backspace at the beginning of the first `<li>`, the cursor jumps to the end of the first `<li>` and deletes the text's last character instead. If that's too "edge case"-y, I'd be glad to approve it since everything else looks great!

obo

jpeterson976:
This seems to be a Slate issue from my testing, for some reason it just doesn't recognize that the cursor has changed position when you click to the start of the line after swapping list types. I'm not sure if that's something we'd be able to accommodate for, so we may just have to leave it be.

zachberry (approved):
Awesome! I think we can accept the edge case that @maufcost discovered, although good catch

PR #1895, opened 2021-08-12
Allow empty XML style tags when saving

### LINKED ISSUE #1654: Empty style tags in XML won't allow the document to save
1. Edit a document in XML
2. Add some text like `<b></b>`
3. Attempt to save (Result: Posting draft failed error)

Empty style tags shouldn't break saving. As far as what to do with them is unclear. In the StyleList data structure StyleRanges of length 0 are deleted. So most likely what we could do is allow `<b></b>` but if the user switches from XML to JSON/Visual Editor and then back to XML those empty styles would be deleted.

jpeterson976:
Fixes #1654

I added a check that sets `node.value = ''` if it doesn't already exist, which prevents the saving error from occurring and lets the empty tag get removed when swapping to the JSON/Visual Editor and back.

### DIFF
--- a/packages/app/obojobo-document-xml-parser/__tests__/src/text-group-parser.test.js
+++ b/packages/app/obojobo-document-xml-parser/__tests__/src/text-group-parser.test.js
@@ -145,4 +145,17 @@ describe('TextGroup parser', () => {
 
 		expect(parsed[0].text.value).toBe('')
 	})
+
+	test('Handles missing text item', () => {
+		setTextGroupElements([
+			{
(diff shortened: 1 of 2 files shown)


maufcost (approved):
LGTM! This works pretty well as far as my tests go. Clap Clap!

I tested by:
* Creating empty <b></b>, <i></i>, and others and going back and forth between editors.
* Styling empty spaces (also in conjunction with other text tags e.g. text, textGroup, etc)

zachberry (approved):
LGTM!

PR #1896, opened 2021-08-16
All other loaded videos pause when one video is playing

walid-i:


jpeterson976 (changes requested):
This works really well in most cases, though if you put multiple of the same video on a page then it'll occasionally not pause other videos if their URL happens to match due to them being at the same timestamp in the video. Comparing the video `id`s instead of their URLs seems to get around this, since each one is given a unique id even if they have the same URL.

zachberry (approved):
Forgive me @walid-i but I added a few last minute tweaks to get this into dev/25. Based on what @jpeterson976 found I changed the code that attempts to pause video from using the video URL that I had originally suggested to the id property that exists on the youtube player object. From what I found that just simply increments each time a video is added to the page and should always be unique.

I also noticed that the loadedVideos array was growing as you changed pages and never got rid of the player objects, so I added some code to remove them from loadedVideos on unmount. To make that slightly easier I changed loadedVideos from an array to an object.

Issue #1644, opened 2021-01-12
Objectives


zachberry:
The desire is to tie a page and/or questions to one or more Module-level objectives.

Some initial ideas in an example XML mockup:

```xml
<?xml version="1.0" encoding="utf-8"?>
<ObojoboDraftDoc>
	<Module title="edX Multiple Choice Question Example">
		<objectives>
			<objective id="644a1848-39bc-4629-8773-3a047eaa2aae">Students will be able to categorize types of animals into the correct classes with a graphic organizer after reading an article on animal traits.</objective>
			<objective id="26b03cca-945f-4b8c-9d15-1b64b882a284">By working in collaborative literature groups, students will be able to form predictions about the next chapter of Anne of Green Gables using at least three pieces of textual evidence.</objective>
		</objectives>
		<Content>
			<Page objective="26b03cca-945f-4b8c-9d15-1b64b882a284">
				<Question>
					<p>test</p>
					<MCAssessment responseType="pick-one">
						<MCChoice score="100">
							<MCAnswer>
								<p>Correct</p>
							</MCAnswer>
							<MCFeedback>
								<p>Correct</p>
							</MCFeedback>
						</MCChoice>
						<MCChoice score="0">
							<MCAnswer>
								<p>Incorrect</p>
							</MCAnswer>
						</MCChoice>
					</MCAssessment>
					<solution>
						<Page>
							<p>This is a description of the solution to the problem.</p>
						</Page>
					</solution>
				</Question>
				<Question objective="644a1848-39bc-4629-8773-3a047eaa2aae">
					<p>test</p>
					<MCAssessment responseType="pick-one">
						<MCChoice score="100">
							<MCAnswer>
								<p>Correct</p>
							</MCAnswer>
							<MCFeedback>
								<p>Correct</p>
							</MCFeedback>
						</MCChoice>
						<MCChoice score="0">
							<MCAnswer>
								<p>Incorrect</p>
							</MCAnswer>
						</MCChoice>
					</MCAssessment>
					<solution>
						<Page>
							<p>This is a description of the solution to the problem.</p>
						</Page>
					</solution>
				</Question>
			</Page>
		</Content>
	</Module>
</ObojoboDraftDoc>
```

My current thoughts:

* Objectives are plain-text only, which makes it technically easier to handle and keeps them simplified
* Objectives are defined only at the Module level. Each objective is given an id (UUID by default).
[...]

PR #1897, opened 2021-08-18
Keidakira/dev

keidakira:
The PR is for Objectives UI, functionality as discusses in Issue #1644


packages/obonode/obojobo-chunks-list/viewer-component.scss
```diff
@@ -6,6 +6,18 @@
...
+			margin: 0.5em 0;
+		}
+
+		div:first-child {
+			margin-top: 0;
+		}
+
+		div:last-child {
+			margin-bottom: 0;
+		}
```
FrenjaminBanklin, line 20: These changes don't seem like they're necessary in the scope of this issue. Is there a specific problem the extra margin is meant to solve?

If not, I recommend removing these rule changes.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/objectives/objective-list-modal.js
```diff
@@ -0,0 +1,207 @@
...
+
+		this.createNewObjective = this.createNewObjective.bind(this)
+		this.handleObjectiveInput = this.handleObjectiveInput.bind(this)
+
+		this.objectiveListHeader = (
+			<div
+				style={{
+					margin: '-0.7em 0em 1em 0em'
+				}}
+			>
```
FrenjaminBanklin, line 39: Add a `className` to this and handle the styling with a rule in the adjacent .scss file.

packages/app/obojobo-document-engine/src/scripts/oboeditor/components/objectives/objective-list-modal.js
```diff
@@ -0,0 +1,207 @@
...
+				done
+				title="Objective Library"
+				onConfirm={() => {
+					this.props.onClose(this.state)
+				}}
+			>
+				{this.state.globalObjectives.length > 0 && this.objectiveListHeader}
+				<div
+					className="objective-list-modal"
+					style={this.state.globalObjectives.length === 0 ? { minHeight: '2.2em' } : null}
```
FrenjaminBanklin, line 176: Recommend adding an additional class if this condition is true, and a corresponding `.scss` rule to control this style.

FrenjaminBanklin (changes requested):
The `.zip` file at `packages/app/obojobo-document-engine/src/scripts/oboeditor/components/objectives.zip` needs to be removed. Ideally it wouldn't be in the commit history at all, but it's small enough that we probably don't need to go that far.

When running tests locally, there are quite a lot of failures in existing tests that need to be accounted for. It looks like the new code is not being tested/covered, either.

Also, it isn't vital, but any file corresponding to a React component/using JSX syntax should probably end in `.jsx` instead of `.js`.

Otherwise some polish and style cleanup as indicated.

PR #1904, opened 2021-09-03
Fixed broken select on Safari

### LINKED ISSUE #1902: Select element in more info box breaks in Safari
It's as if there were a white background partially covering the bottom part of the text inside the `more-info-box`'s select element. My guess is that this issue might be coming from the Sass `@mixin` used on it. 

Screen Shot 2021-09-02 at 12 14 00

maufcost:
Fixes #1902

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.scss
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/more-info-box.scss
@@ -206,8 +206,10 @@
 		border-radius: $dimension-rounded-radius;
 		font-size: 0.7em;
 		background-color: $color-bg;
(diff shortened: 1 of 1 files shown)


maufcost:
It turns out the `select` element didn't have any `@mixins` applied to it. Safari was being stubborn about rendering it with its `appearance` set to `none`. It should look better now. Tested on Chrome, Safari, and Mozilla.

jpeterson976 (approved):
Also tested with the same browsers, select box looks good in all of them!

FrenjaminBanklin (dismissed):
Checked in all three browsers, issue seems to be resolved.

FrenjaminBanklin (approved):

jpeterson976 (approved):

PR #1909, opened 2021-09-14
Invalid assessment attempt modal refreshes the page

### LINKED ISSUE #1657: Change invalidated assessment warning to refresh the page instead of start a new assessment
In dev/16 when you get a warning that an assessment has been invalidated it prompts you to click Restart, which if clicked will put you in an updated assessment. However, if the module has been significantly altered (maybe incorrect information corrected or new content and questions added) that user could be at a disadvantage - they'd now have to answer questions that they haven't seen the content for or having read incorrect information.

I think we change the text slightly on the dialog and change the Refresh button to refresh the page. When that happens the student enters the module fresh as if the invalidated attempt never happened.

I'm open to discussion. We may want to leave the current functionality as is.

jpeterson976:
Resolves #1657 

Modifies the invalid attempt modal to refresh the page instead of starting an attempt, in case the content in the module has changed in any significant way. In the issue Zach said that he's open to discuss it, so if anyone has any ideas about any other/better solutions I'm open to any and all suggestions.

### DIFF
--- a/packages/obonode/obojobo-sections-assessment/components/__snapshots__/assessment-dialog.test.js.snap
+++ b/packages/obonode/obojobo-sections-assessment/components/__snapshots__/assessment-dialog.test.js.snap
@@ -50,7 +50,7 @@ exports[`AssessmentDialog renders as expected AssessmentDialog(assessmentMachine
           Click close dialog to review and catalog your answers.
         </p>
         <p>
-          When you're ready, click restart to start a new attempt with the updated module
+          When you're ready, click reload to view the updated module before starting a new attempt.
         </p>
       </div>
       <div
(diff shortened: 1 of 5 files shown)


FrenjaminBanklin (approved):
Confirmed that attempting to submit an assessment following a change to a module now displays the updated prompt, and follows up by returning to the beginning of the module instead of going straight back to the assessment.

Short of doing a deep comparison to determine exactly which part(s) of the module changed between the time somebody starts an assessment and submits the assessment, I'm not sure what else could really be done here. This is a good enough solution for now.

maufcost (approved):
This looks pretty good to me. I agree with @FrenjaminBanklin above. This should fix any issues we used to have (at least for the time being). Good work :)

PR #1911, opened 2021-09-22
Prevents navigation using buttons when nav is locked

### LINKED ISSUE #775: Navigation via buttons still works even if navigation is locked
If, for example, you have a button that takes you to page 1 that will always work, regardless of if the nav is locked. Then, if there's no button to return or unlock the nav then you're stuck there

jpeterson976:
Fixes #775


maufcost:
Should we be handling cases where: if a page takes you to a page AND locks the navigation. Should we allow this behavior?

obo

FrenjaminBanklin:
Maybe some way of checking the triggers on a page when attempting to add a lock action, and only allowing the lock action if the page has at least one 'unlock' action? But then every time the actions and triggers are adjusted anywhere in a module, we'd have to check the whole thing to make sure the author didn't just create a trap for end users.

I'm not sure there's much we could do to prevent authors from creating a situation like this, and even if there was I think it would be outside the scope of this issue.

maufcost (approved):
This looks pretty good to me as well then. Good job, @jpeterson976 :)

PR #1916, opened 2021-10-05
Added feature to allow authors to restore deleted modules

### LINKED ISSUE #1852: Deleted modules page & Undelete modules
Since modules are soft-deleted we should be able to restore any deleted modules. This feature would be to add a 'deleted modules' page in the Dashboard and give users a way to undelete them with a new API route (ideally multi-select would work in the UI the same as it works currently).

We'll also want to remove the code in #1824 which requires users to type "DELETE" to bulk delete modules. Since recovering the modules will be possible we could revert this back to a standard confirm message.

maufcost:
Fixes #1852


FrenjaminBanklin (approved):
Good job, this'll save a lot of potential support work down the line.

jpeterson976 (approved):
Awesome work, LGTM!

PR #1908, opened 2021-09-14
Remove all Caliper event generation support.

FrenjaminBanklin:
Closes #1907.

**This PR contains a migration.**

Removes all support in the code for generating Caliper events, removes relevant dependencies, and removes the `caliper_store` table.

PR #1919, opened 2021-10-08
Removes Caliper Events

### LINKED ISSUE #1907: Remove Caliper support.
Obojobo will currently create Caliper events matching Obojobo events as those events occur, writing all of the necessary data into the various relevant tables and also writing a row into the `caliper_store` table after extra processing. Currently this data is not being used, and is in fact causing significant database bloat.

In the interest of reducing said bloat and simplifying Obojobo in general, remove all of the current logic for creating Caliper events from the code base. In the future, if having Caliper events becomes necessary, it may make more sense to focus on building out a utility to convert existing Obojobo events into Caliper events on demand rather than storing pre-generated Caliper events in a database table.

iturgeon:
unfortunately we never really ended up using Caliper events in obojobo,
and they are causing our storage needs to double just to store mostly
the same data in another format. This removes all the parts of the code
that creates those events and broadcasts them.

Fixes #1907


FrenjaminBanklin:
It looks like there aren't too many differences between this one and #1908. On top of the things removed in this PR, I also deleted a number of additional files, the Caliper JS dependency, and a few additional references to the `caliper_store` table, which I also created a migration to delete since it will no longer be written to.

For the most part, though, our changes seem pretty much identical. There's some additional cleanup in this PR that would probably improve #1908, so potentially I could see merging the two. If any of the removals in #1908 are potentially destructive to non-Caliper processes (though in testing it seems like everything works fine), feel free to point them out.

PR #1920, opened 2021-10-21
Prevents styling inside HTML nodes

### LINKED ISSUE #1272: Prevent styling inside the HTML chunk

Example:

image

jpeterson976:
Resolves #1272 

This prevents styling within an HTML node from being displayed in the editor, though it technically doesn't prevent them from being applied. To account for this, I also made it so the styling buttons in the header will not show as enabled if the cursor is in an HTML node, since they otherwise might be active when it seems there's no actual styling there.


FrenjaminBanklin:
Tentatively, I think it may still be okay to allow styling inside a Code chunk. It'll allow users with enough free time to highlight text in code examples. Potentially in the (far) future we could embellish the Code chunk to intelligently highlight according to a given language, etc.

FrenjaminBanklin (changes requested):
It looks like this covers most of the styling options, but it's not quite complete.

For example, you can still put links inside HTML chunks, text can be turned into an inline function, and the color of a text can still be 'changed' - although none of these persist in the document's JSON. Inserting links and inline functions do have visible (albeit temporary) effects. 

All of the above illustrated:


Changing the color of the text has no visible effect, however if the cursor is placed inside the area of text where the color was 'changed', you can see the icon change color accordingly.

None of these are really significant bugs, but if possible it would make the most sense to disable all of the styling buttons when the cursor is inside an HTML chunk/when an HTML chunk is selected. If after review this turns out to be a significant undertaking, then we can probably merge this in as-is and maybe spin off a separate issue to see about disabling the buttons.

FrenjaminBanklin:
Closing in favor of #2024.

PR #1923, opened 2021-10-27
Forward delete in empty figure caption no longer deletes the figure

### LINKED ISSUE #1598: Hitting forward delete inside a Figure caption deletes the figure instead of adopting the text below inside the caption
Best seen here - My expectation is that the line "Here is some text" would become the caption for this figure:

Note this is hitting Del and not Backspace.

jpeterson976:
Fixes #1598

### DIFF
--- a/packages/obonode/obojobo-chunks-figure/editor-registration.js
+++ b/packages/obonode/obojobo-chunks-figure/editor-registration.js
@@ -62,16 +62,8 @@ const Figure = {
 			if (event.key === 'Delete') {
 				const text = entry[0].children[0].text
 
-				if (text === '') {
-					const path = Editor.next(editor)[1]
-
+				if (editor.selection.anchor.offset === text.length) {
 					event.preventDefault()
-
-					Transforms.moveNodes(editor, { at: path, to: editor.selection.anchor.path })
-					Transforms.delete(editor, { at: path })
-					Editor.deleteForward(editor)
-
-					Transforms.select(editor, { ...editor.selection.anchor, offset: 0 })
 				}
 			}
 		},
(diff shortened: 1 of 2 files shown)


FrenjaminBanklin (commented):
This definitely solves the problem of forward-delete removing the Figure instead of bringing the next closest text node up into the caption, so on those merits it could be approved and merged in.

However, I'm not sure I agree that this is desirable behavior. If the next chunk happens to be a text type - Heading, Text, Code, List, etc. - it brings the nearest text neighbor up into the caption just fine. However if the next chunk is not one of the above, (Equation, Figure, HTML, IFrame, etc.) it deletes not only the next node but also the Figure itself, as in the original issue.

It may make more sense for forward delete to simply do nothing at all if it reaches the end of an empty Figure caption, rather than lift the next closest text node into the caption in potentially weird and unpredictable ways.

Open to commentary for/against.

jpeterson976:
The non-text node behavior seems like it shouldn't be too difficult to fix, though I think that I agree with you that it might not be desirable. It makes sense for regular text nodes, but figure captions seem like kind of a special case, so I'm leaning towards having it do nothing in all cases.

The default Slate behavior when you're at the end of a non-empty caption seems to have it bring up the next node, but that should be easy enough to prevent so swapping to that behavior should (hopefully) not be too much of an issue.

FrenjaminBanklin (approved):
Forward delete at the end of a Figure caption no longer moves adjacent text nodes or deletes the Figure.

Backspace at the beginning of a Figure caption still has weird behavior in my opinion, but it's probably reasonable in context and would be outside the scope of this issue regardless.

PR #1934, opened 2021-11-22
Standardized clear filter button

### LINKED ISSUE #1930: Standardize the dashboard's 'clear filter' button across browsers.
In Chrome, the button is a dark blue X which is difficult to see against the purple background color of the input field. In Safari, it's a small gray X inside a gray circle. In Firefox, there doesn't appear to be an X at all.

See if it's possible to override the default pseudo-element and use something we can control instead.

maufcost:
I overrode `web-kit`'s UI for our clear filter button within obo's dashboard's search input, and everything seems to be standardized across browsers now.

Fixes #1930

### DIFF
--- a/packages/app/obojobo-repository/shared/components/search.scss
+++ b/packages/app/obojobo-repository/shared/components/search.scss
@@ -21,39 +21,65 @@ $search-input-height: 30px;
 		}
 	}
 
-	input {
-		@include text-input;
+	form {
+		display: flex;
+		align-items: center;
+		justify-content: center;
 
-		z-index: 1;
-		background: none;
-		top: 0;
-		left: 0;
-		width: calc(100% - 2px);
-		line-height: $search-input-height;
-		border: solid 1px $border-color;
-		background-color: white;
-		border-radius: 3px;
-		height: $search-input-height;
-		padding-left: 23px;
-		padding-right: 10px;
-		outline: none;
-		position: relative;
-		-webkit-appearance: none;
+		input {
+			@include text-input;
 
-		&::-webkit-search-cancel-button {
+			z-index: 1;
+			background: none;
+			top: 0;
+			left: 0;
+			width: 100%;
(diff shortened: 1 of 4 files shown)


packages/app/obojobo-repository/shared/components/search.scss
```diff
@@ -41,8 +41,14 @@ $search-input-height: 30px;
 		-webkit-appearance: none;
 
 		&::-webkit-search-cancel-button {
-			color: red;
-			background: red;
+			-webkit-appearance: none;
+
```
FrenjaminBanklin, line 45:
```suggestion
			cursor: pointer;
```

FrenjaminBanklin (changes requested):
This looks good, and it does the job.

Only thing I can think of - now that we have full control over how this looks, add a `cursor: pointer;` to round out the clickable button-ness of it.

It'd be nice if we could also make sure this shows up in Firefox, but it looks like that would require disabling the browser's pseudo-element and constructing our own instead. That will probably have to be its own issue.

packages/app/obojobo-repository/shared/components/search.scss
maufcost: Good call. Just added it.

jpeterson976 (dismissed):
Much more noticeable now on Chrome and Safari. Annoying that it just doesn't show up at all on Firefox, but like Brandon said that could probably become its own separate issue. LGTM!

maufcost:
I just found a solution for Firefox specifically. I will be committing that later today.

maufcost:
Everything seems to be working now on all three browsers. I'm just gonna be converting some measurement units from 'px' to 'em' for consistency now.

FrenjaminBanklin (approved):
Filter input now has a consistent clear button across all browsers.

Great work!

jpeterson976 (approved):
Worked in all 3 browsers, LGTM!

PR #1935, opened 2021-11-23
Dropdown ID inputs for trigger modal

### LINKED ISSUE #1622: Convert id inputs in Triggers modal to drop downs so users can select the page rather than copy/paste ids
It's very easy currently to mistype an obojobo id and when that happens the user is lost when the app crashes or doesn't work in an unhelpful way.

Additionally, we only support one assessment so it probably is annoying to users that they have to create a button to start the assessment and wire it up to the correct assessment when they feel like the system would be smart enough to know.

I'm thinking for inputs that ask for an ID (like, for example, the start assessment action in the Triggers modal, or nav:goto), those could be a `select` with potential targets to select (similar to how the start page option works inside the Module's details menu). If the id was edited manually (such as via XML) and there is no matching id in the system then the control could revert into a text entry box (Ideally with a warning, perhaps part of #1621 ). If the user then manually inputs a matching id (or empties the box), and closes the dialog, then reopens it, it would now be back to a select drop down.

This requires that actions know what valid targets are. For example, nav:goto would only want to show Pages and Assessment sections. the startAttempt action would also only want to show Assessment sections. That needs some brainstorming, but the simple basic version could simply hard code look for obo nodes of a certain type where needed.

Finally as a small nicety, I think if the select box will only have one item (as will be the case for assessments for now), we should automatically select the first item. That way authors won't need to wire up assessment sections manually, however, if creating a button to jump to a page (which at that point the module will likely have more than one page) the select can default to an empty state ("Select one..."), forcing them to choose the page to jump to since we can't make an obvious guess.

jpeterson976:
Closes #1622 

Triggers modal now has dropdowns for actions that take in an id. For ones that go to a page, it'll show the page title similar to selecting the start page for a module, and for `focus:component` it'll give the type of chunk and grab a preview of the text for that chunk if applicable. It'll also revert back to a normal input box if the id that's already there is somehow invalid, and automatically swap back to the dropdown once the entry is either valid or empty.

There were a few node types where getting a text preview was a little less obvious. For math equations it just shows the raw latex, lists will show text from only the first bullet point, and for tables it will only look at text in the top left cell. If anyone has any potentially better ideas I'm definitely open to changing how it works for these, or changing what's shown in the labels altogether if need be.

Screen Shot 2021-11-23 at 12 03 01 PM


FrenjaminBanklin:
There are some stability issues being introduced here. It seems to be okay for adding triggers to certain chunks in the editor, but not certain others.

For example: create a new button node, open the triggers menu for that button, then try adding either a new trigger or a new action. Doing either will completely break the editor.

Trying to add triggers to a page will break the editor in the same way.

FrenjaminBanklin:
Fewer breaks, but it looks like buttons are still behaving weirdly.

Create a new button node, then change the 'Then' for any action to 'Focus on a specific item'. Same breakage as before.

The behavior is a little weird with trying to add 'Focus on a specific item' triggers to pages or to the module itself. Options in the drop-down all show as 'hidden'. I'm guessing they correspond to the available pages, but it's unclear - and also buggy in the viewer, though I suspect that's always been the case. Not sure what to do there - either disable the 'Focus on a specific item' actions for page triggers or just let it ride for now.

jpeterson976:
I updated the way that this works for nav items a bit. For the `focus:component` actions it shows the actual information for the nodes on that page now, and if adding the action to the module itself it'll show the nodes for whatever the starting page is set to, even if it's not the current page.

maufcost:
I didn't get to actually test your changes, but Obojobo breaks when you create a new module and access it. I can only access the modules that I already had locally. New ones break.

The error stack starts at:
`Uncaught TypeError: Cannot read properties of undefined (reading 'children')` on `editor-nav.js`

maufcost (changes requested):
Great job so far! 

I just ran into a quick issue:

obo

I created a bunch of nodes, but the trigger options didn't update. It still has the old nodes before I added the new ones.

jpeterson976:
Should be ready for review now. With the way things are passed around it took me a while to figure it out, but the trigger modals in the nav should be properly updating with new nodes.

I also got it to work with the triggers for the module itself. It'll show the elements for whatever the start page is, even if it's not the currently active page. But for some reason the model with all the information doesn't update with new pages until a refresh, so if the start page is set to a newly created page it will just revert back to the text ID input.

maufcost (dismissed):
This is working pretty well! Great job, @jpeterson976 :) The triggers are all applied correctly as far as my testing goes and the nodes are updated according to the current selected page.

The dropdowns also become regular text inputs when one changes the ids manually on the XML and JSON editors, which will make it possible for me to add validation to those fields after this has been merged in.

FrenjaminBanklin (changes requested):
New action triggers default to `nav:goto`, which is fine, but the default action trigger appears to be incompatible with the optional dropdown because it [lacks an 'id' property](https://github.com/jpeterson976/Obojobo/blob/431ef27df2b619fa65938827a8d52bd0072ae0aa/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/triggers/trigger-list-modal.js#L176).

This should be a pretty quick and easy fix. You've already mentioned the issues when setting up `focus:component` triggers on the module as a whole with regards to changing the starting page, so we could potentially let it ride - but as it is, it's pretty easy for authors to get themselves into a situation where their content doesn't work and they're not sure why. If possible we might need to add some logic on the viewer side to prevent crashes in the event of a `focus:component` event trying to target an item that's not on the current page. That could also be its own issue, maybe.

maufcost:
Just like before: amazing work with this, @jpeterson976 ! Now, I am not sure if that was already there before this issue, but I created a new page, added a bunch of nodes to it, and created a trigger on the new page that's supposed to scroll the user to a specific node in the page. However, this action is not working for me on new pages. Only on the first one. Can anyone replicate this?

Here's the trigger configuration I used for the new page:
Screen Shot 2022-03-15 at 15 32 02

jpeterson976:
With some additional testing, I can't seem to recreate the server error on my end. So I'm not sure what's going on there or if there's actually an issue with the code here that's potentially causing it.

The not scrolling on anything other than the first page is definitely a thing, but since it doesn't seem to be related to the changes I've made here I think it can be turned into a new issue on its own.

maufcost (dismissed):
LGTM then! I didn't find any more bugs on my part. Awesome work.

PR #1937, opened 2021-11-29
Remove isomorphic data table rendering from Stats page.

FrenjaminBanklin:
Following the most recent deploy to QA, we discovered that the stats page was not rendering properly - the initial (server-side) render of the module list was not being styled at all, whereas client-side renders of assessment stats were being styled properly.

Believing this is due to the module list being rendered on the server and the pre-generated style tag not being preserved after hydration, I refactored the Stats page to render only a basic container on the server which then generates the module list following an API call. Assessment stats are then rendered as before.

### DIFF
--- a/packages/app/obojobo-repository/server/routes/api.js
+++ b/packages/app/obojobo-repository/server/routes/api.js
@@ -11,7 +11,7 @@ const {
 	requireCurrentDocument,
 	checkValidationRules,
 	check,
-	requireCanViewSystemStats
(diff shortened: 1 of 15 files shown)


maufcost (approved):
LGTM! Awesome work. The stats page is rendered as expected.

Even some styles that were broken before on dev branches are being properly displayed now (e.g. the "Load stats" button is showing correctly now underneath the module results page. Before this PR, it was being displayed with a huge top margin).

PR #1938, opened 2021-11-30
Add functionality to the local dev tools.

### LINKED ISSUE #1932: Make more dev utilities.
The current set of dev utilities is useful, but could be expanded.

For example, it's currently not easy to really test the assessment stats panel in the dashboard because there's no real way of shortcutting the process of logging in and visiting a module as a student. The existing test instructor user(s) can be given ownership over the example module in order to see the assessment results for the single existing test learning, but that's about it.

Off the top of my head, some improvements could be:
- [x] Allow a module to be specified in the dev utils for the student to visit.
- [x] Allow for new students (maybe even new instructors) to be created with a button press.
- [x] Allow for a specific student (or instructor) to be chosen for usage with the various functions (resource selector, login, module visit, etc.).
- [x] Add roles/permissions (such as access to the Stats page).
- [ ] Anything else that would make sense and help with development.

FrenjaminBanklin:
Closes #1932.

- Adds the ability to generate new instructor or learner users on demand.
  - Can optionally provide first/last names.
- Adds the ability to add/remove permissions to users on demand.
- Adds the ability to choose a specific user for the login/picker tools.
- Adds the ability to choose a specific user and module for the preview/view tool.
  - Retains option for enabling score imports.

Apparently the dev tools never had unit tests, so they continue to not have unit tests.

### DIFF
--- a/packages/app/obojobo-express/server/obo_express_dev.js
+++ b/packages/app/obojobo-express/server/obo_express_dev.js
@@ -207,7 +207,7 @@ module.exports = app => {
 								}
 							</form>
 						</li>
-						<li><b>Add permission to user:</b>
+						<li><b>Remove permission from user:</b>
 							<form id='resource-select-form'
 								method='post'
 								action='/dev/util/permission'>
--- a/packages/app/obojobo-express/server/obo_express_dev.js
+++ b/packages/app/obojobo-express/server/obo_express_dev.js
@@ -11,17 +11,26 @@ const { isTrueParam } = require('obojobo-express/server/util/is_true_param')
 const User = require('obojobo-express/server/models/user')
 const DraftSummary = require('obojobo-repository/server/models/draft_summary')
 
+const POSSIBLE_PERMS = [
+	'canViewEditor',
+	'canCreateDrafts',
+	'canDeleteDrafts',
+	'canPreviewDrafts',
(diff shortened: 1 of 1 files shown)


jpeterson976 (changes requested):
This is awesome, especially the option to add permissions so we don't have to manually update the database to be able to view stats.

Only minor thing I noticed is there are two "add permission" options on the page, though the code for the second one actually removes instead of adds. I also had the idea that being able to remove users might be good, though looking at the endpoints there's not actually anything that does that so it'd probably be more work than it's worth and likely wouldn't get much use anyway.

FrenjaminBanklin:
Looks like that was a copy/paste fail. Fixed it and pushed.

It might be interesting to see what breaks if users are deleted, but since that's currently not expected to ever happen in practice. Like you said, there doesn't appear to be anything anywhere in the code to handle deleting user records - if we wanted to do that, we'd have to remove them from the database by hand regardless of the environment. But since it doesn't seem like it's even possible, I think I'm okay with not having a button for it.

jpeterson976 (approved):
Awesome, LGTM.

I didn't think about all the other repercussions of deleting a user and the cleanup that it would require, so that's definitely not something that's necessary to add.

PR #1940, opened 2021-12-02
Added justified alignment option

### LINKED ISSUE #1926: Add a 'Justified' alignment option.
Currently it's possible for text to be left-, center-, or right-aligned, but not [justified](https://www.w3schools.com/cssref/pr_text_text-align.ASP).

maufcost:
Fixes #1926

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
@@ -837,6 +837,23 @@ Ctrl+Shift+R"
         </svg>
       </button>
       <button
+        aria-label="Justify
+Ctrl+Shift+J"
+        className=" is-not-selected"
+        onClick={[Function]}
+        title="Justify
+Ctrl+Shift+J"
+      >
+        <svg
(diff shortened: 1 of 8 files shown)


deundrewilliams (commented):
Tested in each browser, the justification button in the menu works for all, but it looks like the keystroke (cmd+shift+J) only works in Chrome and Safari

maufcost:
Thanks for pointing this out. I will investigate.

SJJacques:
The button and command both work for me on all browsers. It only justifies on Chrome; Firefox and Safari change alignment but there's no difference between left align and justify.

maufcost:
Everything seems to be working now on all major browsers (Chrome, Safari, and Firefox). The issue before seemed to be a `white-space` bug exclusively on Firefox that developers had faced before. It's fixed now. Let me know if anyone comes across any issues.

SJJacques:
Tested it out and everything seems to work.

deundrewilliams (dismissed):
Tested again and everything (menu button and keystroke) works in each browser

jpeterson976 (approved):
Both the button and the command work in all three browsers for me as well. LGTM

jpeterson976 (approved):
Re-approving after base change

SJJacques (approved):
Also approving after base change.

PR #1942, opened 2021-12-07
Accessibility Fixes for Assessment Updates

### LINKED ISSUE #1788: Accessibility tweaks to assessment updates in Numeric branch
Found two issues testing the numeric branch with Chromevox

1. When the assessment dialog changes from Submitting your responses to the results dialog Chromevox stays silent, so you don't know that the state of the dialog changed.
2. There's no announcement when a question in an assessment saves your answer. The only indication that something is happening is that the inputs are read as disabled, since the question transitions into a disabled state while saving.

deundrewilliams:
Fixes #1788 
* Adds `aria-live` to the modal component so the screenreader will read any text updates inside, this affects the response dialog after submitting an assessment
* Adds a response status cache and `aria-live` to `QuestionResponseStatus` to keep track of statuses already read + avoid repeating announcements. Whenever a new status is received, text inside a hidden div is updated and read by the screenreader


FrenjaminBanklin (approved):
Checked this with Chromevox. It looks like it addresses the shortcomings mentioned in the ticket - the screen reader indicates when an answer is saved and when a submission has finished. Well done!

I did notice while testing this that using the keyboard to navigate multiple choice questions with only a single correct answer is extremely difficult, but since that also appears to be happening on the current dev branch I'm going to say it isn't due to your changes. I'll open a separate issue.

FrenjaminBanklin (approved):
I could see making tweaks to what specifically is read out in the future, but this does a good job of addressing the problem for now.

PR #1944, opened 2022-01-06
Adopted the new settings dialog component for obo nodes

### LINKED ISSUE #1775: Editor Settings Dialog
Adopt the settings dialog used by the Materia chunk for all the other chunks that wouldn't require radical customization.

maufcost:
**Node dialogs that were not updated:**
- text, heading, list, horizontal line, code, question, question bank, html: no settings dialog.
- math, button, and table: uses a different type of dialog.
- iframe: not updated because of #1828.
- Materia widget: already updated.

**Node dialogs that were updated:**
- figure: both dialogs updated.
- Youtube: updated.

As always, let me know if you would like me to make any changes.

Fixes #1775


jpeterson976 (changes requested):
The only thing that I noticed is that the width and height input boxes for custom sized figures are no longer centered vertically with the new dialog, but everything else looks to be the same as the old versions.

maufcost:
Good catch! Fixing that now.

jpeterson976 (approved):
Great, LGTM!

deundrewilliams (dismissed):
All updated dialogs are centered, lgtm

FrenjaminBanklin:
Everything is functioning properly after the change, so good work there.

The only thing I can see possibly still needing work (unless this is somehow only happening to me) is that the settings windows seem to have way more white space than they need. For example, currently running on production:
Screen Shot 2022-01-27 at 3 06 57 PM
versus:
Screen Shot 2022-01-27 at 3 07 08 PM
and:
Screen Shot 2022-01-27 at 3 08 45 PM
versus:
Screen Shot 2022-01-27 at 3 09 13 PM

If it's a significant burden to tighten these up then I wouldn't necessarily block the merge over it, but if it's possible then I think it would be preferred over the significant extra padding.

FrenjaminBanklin:
@maufcost Double-check that the two minor changes I made to the CSS didn't bug up anything you've done, otherwise this looks like it's ready to go.

maufcost:
Checking that now, and I will get back to you in a couple minutes :)

maufcost:
Just checked all the updated modals. As far as my testing goes, they are good to go!

deundrewilliams (approved):
Dialogs are still centered after CSS updates, lgtm

jpeterson976 (dismissed):
Stuff still works, LGTM!

PR #1952, opened 2022-01-21
Allows nav:goto to ignore navigation locks by default, adds a new setting to enforce navigation lock per action.

### LINKED ISSUE #1948: Restore the ability for button-based navigation to ignore navigation locks.
Reverses the solution for #775.

It's easy for content authors to create situations in which end users can become stuck in a certain page due to navigation being locked. Previously, `nav:goto` triggers were able to bypass navigation locks which, in some cases, would cause problems and, in other cases, solve problems.

In practice, that behavior ended up becoming load-bearing. Existing content can now be assumed to have the expectation that buttons using the `nav:goto` trigger will assume that navigation will still work regardless of being locked.

Either rewind the fix and allow `nav:goto` triggers to ignore navigation locks again, or embellish such events with an extra configurable property which makes it explicit that the buttons will respect any locks on the navigation.

FrenjaminBanklin:
Closes #1948

Adds an 'ignore lock' toggle to nav:goto actions which defaults to true, allowing all existing button-based navigation to continue working while also allowing authors to respect navlocks, or at least inform them when they do not.


deundrewilliams (commented):
The ignore navigation toggle works, allowing navigation when it's switched on, but it only works with 'Go to' and not 'Go to previous page' or 'Go to next page'. Should the toggle be added to these options, or is it fine to not have it since the user can just enter the ID of the next/prev page in the field?

FrenjaminBanklin:
@jpeterson976 worked on this more recently so he may remember if the 'Next Page' and 'Previous Page' buttons still worked regardless of the navigation being locked, but I don't think they did. If not, it would be inconsistent to allow a triggered `nav:prev` or `nav:next` action to optionally ignore the navigation lock.

It wouldn't be hard to enable the new optional behavior for `nav:prev` and `nav:next`, and I thought about doing so, but I think it makes sense for those two to respect navigation locks at all times based on the above assumption. With `nav:goto` I can think of situations where authors want the nav to stay locked but for specific situations to still move students around in the content.

deundrewilliams (approved):

jpeterson976 (approved):
I checked my old PR, `nav:prev` and `nav:next` did not previously work when navigation was locked, so Brandon's assumption was correct.

The new toggle switch also works as advertised, so this gets my approval.

PR #1954, opened 2022-01-25
Add CDN asset hosting option

iturgeon:
Using a CDN for the compiled assets will free up the node server to focus on the tasks it's really needed for.  Doing so will speed up the server, reduce server hosting costs, and speed up end user's load times.

Since this added 2 env vars, I figured I'd add a list on configuration env vars to the readme.

Adds:
- CDN_ASSET_HOST (optional)
- TMP_UPLOAD_DIR (required - unless you change media.json)

To use a CDN, set CDN_ASSET_HOST=https://cdn.com
Then upload everything from packages/app/obojobo-express/server/public to the CDN after running `yarn build` for the exact version of the site you deploy.  

To add further detail:  `yarn build` uses webpack, and the assets are named with a  name that includes a hash based on the content of the file.  Webpack creates a manifest file that is used to look up this hash name - if obojobo is looking for 'fancy-file.js' the manifest is used to look up the built file's name `fancy-file-<hash>.js`.  Because of this - the files on the CDN need to be updated when the js/css/images are updated because the updated build for obojobo will be looking for a new file by name.  

This means the deploy process would need to include running yarn build then uploading the public folder to the CDN on every deploy of a new version of the application.


iturgeon:
UPDATE: The latest commit will allow module images to be loaded through a CDN as well.  

This would require the CDN to be able to pass requests to `/library/module-icon/*` to obojobo server so the server can still build them, but the CDN can cache them

PR #1964, opened 2022-01-29
adds readOnly to dashboard module checkbox controls

iturgeon:
React complains about readOnly not being set on this checkbox when it's rendered.


FrenjaminBanklin:
Given how low-impact this is, and since it looks like it's working as intended, I'm going to go ahead and merge it in.

PR #1967, opened 2022-02-03
Moved most test files to be adjacent to the files they test

SJJacques:
So it turns out the file paths work out for most of these by default when the test files are moved because the layout of the `__tests__` folder matches that of the `src` folder in the same directory. I moved everything except for some tests in `obojobo-document-json-parser` and `obojobo-document-json-parser` because they're a little different. The tests work and pass, but I have not exhaustively scrutinized them.


SJJacques:
I think the conflicts exist because I've removed the `__tests__` directory from the packages I could find the adjacent files for.

PR #1969, opened 2022-02-08
More info box z-index issue in firefox

### LINKED ISSUE #1950: More info box z-index is not respected in Firefox when something is selected in slate
The slate text appears above the more info box as shown below.

Screen Shot 2022-01-21 at 11 01 37 AM

jpeterson976:
Fixes #1950 

Adjusted `&.is-selected` to just `.is-selected` in the CSS for nodes, which for whatever CSS reason seems to have fixed things appearing on top of the more info box.

### DIFF
--- a/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/editor-nav.scss
+++ b/packages/app/obojobo-document-engine/src/scripts/oboeditor/components/navigation/editor-nav.scss
@@ -8,7 +8,7 @@
 
 	height: inherit;
 	display: table-cell;
-	z-index: $z-index-above-content;
+	z-index: $z-index-above-most;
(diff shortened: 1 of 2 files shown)


FrenjaminBanklin (changes requested):
This seems to work for simple chunks, but it only changes the layer that causes the problem. Any chunk that has sub-components which themselves can have the `is-selected` class will still exhibit the behavior. For example, a Figure:


Odds are the actual problem is that the `z-index` of the `.oboeditor-component` element is higher than some ancestor of the `.more-info-box` element - possibly the whole nav bar itself just needs a higher z-index, but you'd have to experiment to make sure there aren't any gotchas by doing that.

maufcost (approved):
Sweet! LGTM!

PR #1972, opened 2022-02-10
Implemented logic to stop duplicate attempts to be submitted to grade book

### LINKED ISSUE #1962: Enable additional score passback control.
In some LMSes (Canvas, for example) it's possible for assignments to be submitted past a given due date, which in some cases may incur optional late submission penalties.

Currently, every assessment attempt with a passing score will submit the highest score to the gradebook, whether the score is equal to or lower than the previous highest score. In cases where a learner has previously submitted the highest possible score and is attempting an assessment again for reference or practice, submitting another passing attempt (regardless of score) could potentially lower their score in the LMS gradebook due to late submission penalties.

It is possible to configure an Obojobo assessment to only send a score to the gradebook if it meets a certain threshold. Investigate whether it's possible to add logic that will only send a score to the gradebook if it's higher than the previous highest score. Alternatively, allow the student to choose whether or not to resubmit their score.

maufcost:
Fixes #1962


FrenjaminBanklin:
I think it may still be valuable to store a record in the `lti_assessment_scores` table even if we don't actually send a score back to the LMS - for example, when somebody completes an assessment during a preview we still write a log to that table with a `status` value of 'not_attempted_preview_mode' and a `gradebook_status` value of 'ok_preview_mode', as seen [here](https://github.com/ucfopen/Obojobo/blob/dev/28-jadeite/packages/app/obojobo-express/server/lti.js#L570).

I'm not 100% familiar with the entire post-assessment process, but generally speaking we'd have a new `status` value and a new `gradebook_status` value to indicate these situations - we would store all of the relevant data in the `lti_assessment_scores` table using these new values, and the process would otherwise be identical to the normal one with the exception that it doesn't actually send anything back to the LMS.

maufcost:
Got it. That works.

I will create a new `status` and `gradebook_status` values that won't send anything back to the LMS. Beforehand, I assumed that anything on `lti_assessment_scores` was being sent to the LMS. After your last comment, I believe for a score in `lti_assessment_scores` to be sent to the LMS, its status field should be set to `success` and the `gradebook_status` set to `ok_gradebook_matches_assessment_score`. I will make the proper changes.

FrenjaminBanklin (changes requested):
Server logs indicate database errors when trying to insert a new row in `lti_assessment_scores` with a `status` value of 'ok_not_considered' due to that value not being in the allowable list. You can see what the list is and how the list was created in [the migration that created the table](https://github.com/ucfopen/Obojobo/blob/95e859de369d8f4cb3d1386393d38189b5f595ad/packages/obonode/obojobo-sections-assessment/migrations/20180104161926-create-lti-assessment-scores.js#L35-L47).

This will require a migration to add 'ok_not_considered' to the list of allowable values. You can read about that query [here](https://www.postgresql.org/docs/current/sql-altertype.html). While the `up` half of this migration shouldn't be too much trouble, there are some problems with the potential `down` - namely that while it looks like it's pretty easy to add values to an enum, it's not even possible to remove values from an enum. Quick googling suggests that the best/only way of doing it is to [create a new one, swap them, and delete the old one](https://stackoverflow.com/questions/25811017/how-to-delete-an-enum-type-value-in-postgres). On top of that, you'd probably have to handle rows which contain the new value since, if you have to migrate down, those rows would no longer be valid.

maufcost:
I might be releasing this issue soon. Adding changes to the db schema was pretty straightforward. However, not being able to change those enum fields has proven cumbersome considering that by the time we have to change a specific row, new rows will have already been added to the database (which may lead to all sorts of problems with row ordering in the db).

maufcost:
Just updated the migration to allow for the new grade book status. I don't think we will need to deal with changing those enums and fall into that huge conundrum of "swapping rows" since the scores are supposed to be updated as new scores come up on the database. Therefore, at least on my end, all server errors and warnings vanished now.

Screen Shot 2022-03-15 at 14 48 36

The most important fields to note here are: `assessment_score_id`, `score_sent`, `gradebook_status` fields: the `gradebook_status` will only have an `ok_matches_assessment_score` if the score is higher than the score already on the grade book. Otherwise, the score will not be considered (thus resolving that problem of students taking assessments for practice after the due date and getting their grades penalized). The `assessment_score_id` field is also updated according to whatever score in the `assessment_scores` table is used in the grade book.

FrenjaminBanklin:
Adjusting the existing migration to add the new enum options will only work for anybody starting up a new instance of Obojobo from scratch. Existing instances will not run that migration again, they would have to manually run some kind of query to alter the list of allowed values if there isn't a new migration to do it for them. This'll have to be a new migration with its own `up` and `down` - the `up` would add the new option to the existing enum and the `down` would probably have to go through the whole copy-swap-delete-rename process described in the link I posted earlier.

It might make sense to go further than just using the `gradebook_status` column to track this scenario. Specifically I think the `score_sent` column is a bit disingenuous here since we're specifically not sending anything back to the gradebook. We're recording the assessment score elsewhere and referring to that record with the `assessment_score_id` column in this table, so I'm thinking maybe we should set `score_sent` to `0` or `NULL` for these records in addition to 'ok_not_considered' in `gradebook_status`. This way, even in the situation where we have to remove 'ok_not_considered' as a valid option for the enum, instead of deleting about-to-be-invalid rows we could set their `gradebook_status` value to 'ok_gradebook_matches_assessment_score' and still be able to tell them apart by the absence of a value in `score_sent`.

PR #1986, opened 2022-03-04
Fixes numeric/MC question survey toggle misbehavior.

### LINKED ISSUE #1922: Fix 'Survey Only' toggle behavior in the editor for numeric questions.
Toggling 'Survey Only' on/off has weird effects on the editor with numeric input questions.

Example 1:
Create a new question. Click 'Survey Only'. Change the question type to 'Input a number'. Though the correct/incorrect toggle does not appear (as expected for survey questions), the answer and answer type inputs are still available.

Example 2:
Create a new question. Change the question type to 'Input a number'. Click 'Survey Only'. The entire answer disappears.

In both cases, after saving the module and then previewing, the question still looks and acts like a numeric input question. Providing any number yields a 'Response recorded' (as expected for survey questions), but even if providing the correct value as determined in the editor, feedback does not appear like it does for multiple choice questions.

There could be situations where it makes sense to have a numeric question that is treated like a survey question, but in those cases the feedback should probably still appear when appropriate values are provided as 'answers', and the editor needs to behave the same way it does for multiple choice questions.

FrenjaminBanklin:
Fixes #1922.

Makes sure `questionType`, whether 'default' or 'survey', is properly applied when changing a question's type between multiple choice or numeric.

### DIFF
--- a/packages/obonode/obojobo-chunks-numeric-assessment/editor-component.js
+++ b/packages/obonode/obojobo-chunks-numeric-assessment/editor-component.js
@@ -40,7 +40,7 @@ class NumericAssessment extends React.Component {
 			<div
 				className={
 					'component obojobo-draft--chunks--numeric-assessment is-type-' +
-					this.props.element.questionType
+					(this.props.element.questionType || 'default')
 				}
 			>
 				{this.props.children}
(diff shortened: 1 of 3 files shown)


maufcost (dismissed):
LGTM! I tested the two cases you posted on this PR's original issue, and both of them seem to be sorted out. Great job!

jpeterson976 (dismissed):
^ Same, both cases seem to work properly now. LGTM!

FrenjaminBanklin:
This hasn't changed since the previous approval reviews. Going to go ahead and merge it.

PR #1988, opened 2022-03-15
Keidakira/dev

keidakira:
A branch for the Objectives issue (https://github.com/ucfopen/Obojobo/issues/1644) #1644


FrenjaminBanklin:
Yeah, I think that would be the better of the two options.

It may also make sense to close existing PRs as needed, depending on what is no longer relevant.

keidakira:
Sure, I'll close the PRs. Thanks for helping me out!

PR #1989, opened 2022-03-17
Issue/1644 objectives

keidakira:
A PR for the objectives issue #1644 . Please have a look. Thank you!


FrenjaminBanklin:
I'm confused by the diff in this pull request, since it looks like a lot of these "changes" are already present in the dev/28 branch. Quite a few of them even appear to be from previous releases, so I have no idea why they're showing up as new changes here.

It looks like this is actually the same source branch as the previous pull requests, so I assume these weird issues are the result of merge conflict resolutions that went off the rails.

I'll clarify my suggestion from earlier. I would recommend checking out the current dev branch, `dev/28-jadeite`, then **create a new branch** (e.g. `git checkout -b`) from that branch, then copy **only the relevant changes for this issue** into the new branch. As it is, I can't tell which changes were made for this specific issue and which changes are erroneous.

Starting with a fresh branch from the current dev commit that only has changes relevant to the objectives features should hopefully help avoid the issues we're seeing here and make everything easier to review.

keidakira:
I think it might be because I copy paste few folders to make it work. So I should only modify the changes that are required in files and not copy paste any folders right?

FrenjaminBanklin:
I would generally advise against copying directories from the old branch, yes, but you should be cautious even when copying files as there's always the possibility that you might overwrite something that has been added in a more recent release.

The idea is that if you branch from `dev/28-jadeite` and remake only the relevant changes, there will be fewer (or, if we're lucky, zero) conflicts between your work and others' work that has been released in the meantime.

keidakira:
I see, then I'll redo the PR. Thank you for explaining it patiently.

PR #1994, opened 2022-04-07
adds bulk downloading, sharing, and copying

walid-i:


FrenjaminBanklin:
The introduction of the private collection changes will probably complicate these merge conflicts, so I'll take a look at them and see about working out resolutions.

FrenjaminBanklin:
In addition to the merge conflicts (this was branched from a few releases back and hasn't been refreshed with recent updates in the meantime), it looks like there are numerous issues in the new functionality. The multi-select count is way off, which breaks bulk downloads, and bulk sharing looks unfinished - the module permissions dialog is still rendering as if it only has a single module selected, and the new call in the dashboard component doesn't provide a module, but instead provides the whole props object for the dashboard.

While merging the latest dev branch into this branch, I've resolved all of the merge conflicts and somehow resolved the multi-select count mismatch. Bulk copying was tweaked a little to work without changing how every other bulk function works. While fixes are in progress for the remaining issues, I'm converting this to a draft.

PR #1997, opened 2022-04-19
Draft: Add Access Levels to Sharing Permissions

### LINKED ISSUE #1983: Fix Share and Delete Permissions for Modules
Currently, if an instructor (Instructor 1) creates a module and shares it with someone else (Instructor 2), Instructor 2 is able to share and edit that module. The option to delete appears for Instructor 2, but clicking the button just closes the modal without deleting anything. Since Instructor 2 can share that module, they can also take away anyone's access, including Instructor 1. If Instructor 1 no longer has access, that module can't be deleted by anyone, since currently a module can only be deleted by its creator. So, there seems to be at least two options:

1. Don't allow users to change sharing permissions of a module that they don't own, and disable/hide the sharing and deleting buttons for these users. This could also just be to don't allow users to remove access from the module owner.
2. Allow any user to delete any module that is shared with them

deundrewilliams:
**_This PR has a migration_**

(fixes #1983)

This adds different access levels (Full, Partial, and Minimal) to modules. When sharing, the user will have the option of giving them Full, Partial, or Minimal access.

Full: All options
Partial: View stats, copy, edit, preview
Minimal: View stats, copy, preview

**Changes:**
All existing modules will give every user that already has access Full level permissions.

Previously, anyone with access had the option to delete a module, but this would only work for the module creator, and in every other instance, this would fail silently. Now, anyone with Full level access can delete and restore the module. If a module was deleted but you didn't have Full level access to it, it won't appear in your Deleted Modules, but it will re-appear in your modules if it's restored by someone with Full access.

When sharing a module, the default access level will be Full, but this can be changed afterwards. Also, a user cannot change their own access level, this is to prevent the situation of a module not having anyone with Full access.
Screen Shot 2022-04-19 at 12 23 01 PM

When showing the module menus, only the actions that a user has access to for each module will appear.
* Minimal:
Minimal module menu 
Minimal module options

* Partial:
Partial module menu
Partial module options

* Full:
Full module menu
Screen Shot 2022-04-19 at 12 17 36 PM

In the editor, the 'Delete' button is disabled for users with a Partial access level.
Screen Shot 2022-04-19 at 12 26 32 PM

When selecting modules for bulk operations, if a certain operation isn't available for all selected items (because of the user's access level), that button will be disabled.

Screen Shot 2022-04-19 at 12 30 31 PM

Screen Shot 2022-04-19 at 12 30 38 PM

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/__snapshots__/visual-editor.test.js.snap
@@ -534,6 +534,7 @@ exports[`VisualEditor VisualEditor component 1`] = `
       </button>
     </div>
     <div
+      accessLevel="Full"
       className="mockFileToolbarViewer"
       insertableItems="mock-insertable-items"
       mode="visual"
@@ -1177,6 +1178,7 @@ exports[`VisualEditor VisualEditor component changes pages 1`] = `
       </button>
     </div>
     <div
+      accessLevel="Full"
       className="mockFileToolbarViewer"
       insertableItems="mock-insertable-items"
       mode="visual"
@@ -1820,6 +1822,7 @@ exports[`VisualEditor VisualEditor component changes pages 2`] = `
       </button>
     </div>
     <div
+      accessLevel="Full"
       className="mockFileToolbarViewer"
       insertableItems="mock-insertable-items"
       mode="visual"
@@ -2509,6 +2512,7 @@ exports[`VisualEditor VisualEditor component with no page 1`] = `
       </button>
     </div>
     <div
+      accessLevel="Full"
       className="mockFileToolbarViewer"
       insertableItems="mock-insertable-items"
       mode="visual"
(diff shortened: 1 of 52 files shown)


FrenjaminBanklin (changes requested):
This is a great addition and everything seems to be functioning well. I can only see a few things that I might want to be changed.

1. The spacing for individual user items in the Module Access dialog looks strange, but I'm not sure how best to address it. My first thought is to left-align everything and line up the avatar with the name:
image
I'm open to alternatives. It might be nice to tighten up the vertical padding in the access level drop-down so each item takes up less vertical space, but I'm not sure how well that'll work or if the few pixels it would save are worth it.

2. Instead of storing the full text description of access levels in the database (i.e. 'Full', 'Partial', 'Minimal') it would be preferable to map those permission levels to numeric constants for storage purposes (i.e. 'Full' => 1, 'Partial' => 2', 'Minimal' => 3, etc.). Since the frontend and backend use the same language, those constants could be defined in a single place and shared as necessary.

FrenjaminBanklin (approved):
Everything seems to be working well and looks good. I'll call this one ready to go.

Two things I've noticed that can be spun off into their own issue:
 - What changes based on the permission level is not immediately evident. We could maybe put a tooltip somewhere in the upper half of the interface explaining what each level allows.
 - If a user has full access to a module when they visit their dashboard and do not refresh their dashboard after their access to that module is changed, they will still be able to interact with that module as if they have full access. I don't see this as too big a deal, but it would be nice to have some additional checks in the backend to invalidate any actions outside of a user's current access level.

PR #1999, opened 2022-04-20
Makes "All Modules" button always visible

### LINKED ISSUE #1998: Deleted modules are unreachable if a user has less than 6 total (not deleted) modules
Since the "All Modules" button on the Dashboard only appears if the user has more than 5 recent modules, any user with less than this amount can't get to the All Modules page and see their deleted modules/restore these modules (other than manually adding /all to the end of the dashboard URL).
Screen Shot 2022-04-20 at 9 26 17 AM

deundrewilliams:
This allows users with less than 6 modules to be able to reach the All Modules page and get to their deleted modules
Screen Shot 2022-04-20 at 10 07 01 AM

(fixes #1998)

### DIFF
--- a/packages/app/obojobo-repository/shared/components/__snapshots__/dashboard.test.js.snap
+++ b/packages/app/obojobo-repository/shared/components/__snapshots__/dashboard.test.js.snap
@@ -387,6 +387,9 @@ exports[`Dashboard renders with default props 1`] = `
                   New Module
                 </mock-Button>
               </p>
+              <mock-ButtonLink>
+                All Modules
+              </mock-ButtonLink>
             </div>
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


FrenjaminBanklin (approved):
In the short term this will certainly solve the problem. I guess this scenario hadn't crossed my mind since restoring deleted modules wasn't possible when these dashboard changes were first created.

In the long term, I think a better approach might be to have a separate step for detecting when a user has deleted modules, and basing the presence of the button on that instead of just always making it available in the 'All' mode. That can be its own issue, though.

maufcost (approved):
Works fine to me as well, and I agree with @FrenjaminBanklin above. Good job!

PR #2016, opened 2022-07-10
Adds embedded figure binaries in json upload

iturgeon:
Adds initial support for uploading obojobo module .json files
that can include image binary data in the json.  Previously it
was not possible to import a json file with the images referenced
in it's chunks due to them referencing ID's in specific obojobo
instance they were being created on.

This change allows figure chunks to support a new content property
named 'imageBinary' that contains base64 encoded image data when
uploading via the dashboard in .json formatted modules.

This also provides a script that can be run against an existing
.json module on your local files and populate imageBinary for you.

It will will attempt to locate images by their content.filename
in the directory (or sub directory) of the .json module you provide

See embedFigureBinary.js for usage.

This change does not support embedding image binaries in xml formatted
modules and I have not tested support for uploading .json in the editor.

(cherry picked from commit 9cc4a4f7401709417807b5543effc031900ec315)


iturgeon:
The tests don't currently pass & I'm not sure exactly which branch to target, you'll have to let me know @FrenjaminBanklin

iturgeon:
To upload a json file with curl,

```
curl 'https://127.0.0.1:8080/api/drafts/new' \
  --compressed \
  --insecure \
-X 'POST' \
-H 'content-type: application/json' \
-H 'accept: application/json' \
-H 'Cookie: obo_next=SESSIONID' \
-d @<(jq '{"format": "application/json", "content": .}' < ./path/to/lo.json)
```

PR #2019, opened 2022-07-25
Add support for x-forwarded-for with multiple ips

iturgeon:
x-forwarded-for can contain a chain of multiple ip addresses if there are multiple proxies. That can cause errors when attempting to insert into the events table.  This minor change will pick the first value if there are multiple.


FrenjaminBanklin:
@iturgeon This looks fairly straightforward, but I'm not sure how to check this locally to make sure it's working in practice. Regardless of the value of the `x-forwarded-for` request header it doesn't seem like `req.connection.remoteAddress` ever changes. Is there something I'm missing?

FrenjaminBanklin (approved):
Not really sure how to test this properly, but it seems straightforward and doesn't look like it'll introduce any issues.

PR #2020, opened 2022-08-01
Partial scoring for pick-all questions

### LINKED ISSUE #545: Partial scores for MC Questions
_Friday Jun 15, 2018 at 15:10 GMT_
_Originally opened as https://github.com/ucfcdl/Obojobo-Document-Engine/issues/249_

----

Currently MCChoice items MUST have a score of `0` or `100`. Would like to be able to implement partial scores for partially correct answers.

jpeterson976:
Resolves #545 

MC questions that are pick-all should now calculate their scores based on how many correct/incorrect options were chosen, instead of being all-or-nothing.

### DIFF
--- a/packages/obonode/obojobo-chunks-multiple-choice-assessment/server/mcassessment.js
+++ b/packages/obonode/obojobo-chunks-multiple-choice-assessment/server/mcassessment.js
@@ -24,12 +24,17 @@ class MCAssessment extends DraftNode {
 
 				const responseIds = new Set(responseRecord.response.ids)
 
-				if (correctIds.size !== responseIds.size) return setScore(0)
+				let score,
+					numCorrect = 0
 
-				let score = 100
-				correctIds.forEach(id => {
-					if (!responseIds.has(id)) score = 0
+				responseIds.forEach(id => {
+					if (correctIds.has(id)) numCorrect++
+					else numCorrect--
 				})
+
+				if (numCorrect <= 0) score = 0
+				else score = (100 * numCorrect) / correctIds.size
+
 				setScore(score)
 				break
 			}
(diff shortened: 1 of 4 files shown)


FrenjaminBanklin:
While I see this as a clear improvement, it does change the way assessments are graded which could have unintended consequences on grade review.

For example if a student took an assessment before this update and got no credit for a 'select all that apply' question, then go back to review that assessment score after this update goes into effect, do they still see the 0% for that question, or does the partial credit calculation say they should have received credit?
I also assume this means if a student in the above situation goes to an assessment where before they lost points on select-all questions and answers those questions the exact same way, they'd end up getting more points than their previous attempts.

I can also imagine a scenario where authors want a multi-select question that specifically grants a 0% if the student does not select all possible correct answers. Is that readily possible? If not, MC questions would need some kind of a toggle which defaults to the old behavior so everything stays backwards compatible.

jpeterson976:
It looks like previous attempt scores are correctly showing 0% for a partially incorrect question, if the assessment was taken without partial scores.

That's a good point about allowing authors to choose whether they want partial scoring or not though, I'll work on adding a toggle for it.

FrenjaminBanklin (approved):
I couldn't find any gotchas, and everything looks good and seems to be working well. Approved based on that.

I added a last-minute bit of logic to only show the 'partial credit' toggle when the question is a MC type and the 'choose all correct answers' option is chosen since it didn't make sense to make a whole new issue. Little awkward approving my own work, so another set of eyes on it just to make sure I didn't miss anything would be nice.

RachelDau (approved):
I checkout this out and it all looks good on all browsers. Button to toggle partial scoring only appears when it is suppost to and works well. Scoring looks good and the code is clean.

RachelDau (approved):
LGTM

RachelDau:
I spent some time creating different scenarios in assessments and manually checking the math of the score and it looks good. I didn't find any issues with the scoring.

PR #2023, opened 2022-08-26
WIP: Dev/29 - Sodalite

### LINKED ISSUE #1272: Prevent styling inside the HTML chunk

### LINKED ISSUE #1644: Objectives

### LINKED ISSUE #1770: Dashboard module selected state could use a better outline
After changing to the new module styles, the selected outline got a little weird.  It looks like the white outline that should be around just the gem is around the whole block.  this may require some creative thought. perhaps an overlay or a background image behind the gem that's slightly larger 🤔 

Screen Shot 2021-04-09 at 3 13 31 PM

### LINKED ISSUE #1788: Accessibility tweaks to assessment updates in Numeric branch

### LINKED ISSUE #1922: Fix 'Survey Only' toggle behavior in the editor for numeric questions.

### LINKED ISSUE #1983: Fix Share and Delete Permissions for Modules

### LINKED ISSUE #2012: Numeric questions with an answer of '0' can not have a percentage margin of error.
It looks like there's a check in the viewer for this scenario (which throws a fatal error which results in a white screen), but there's no check in the editor to prevent it.

Speculative fixes:
1. Allow this in the editor, change the error detection in the viewer so it doesn't break everything when it happens.
2. Whenever the answer type, correct answer, or margin of error type are changed on a numeric answer, check all three - if the correct answer is zero and the margin of error type is 'Percent', put some kind of indicator in the editor to point out that this is a problem.
    a. This should also prevent the module from being saved, and probably also create an `alert` pointing out the problem so authors don't miss it.
3. Whenever the correct answer of a numeric question is zero, remove the 'Percent' option from the 'Error Type' dropdown. If the 'Error Type' value is currently set to 'Percent', it should be automatically set to 'Absolute' - make sure this is reflected accurately in the JSON/XML for the module.

### LINKED ISSUE #2026: YouTube videos have extra margins in question feedback and explanation sections.
YouTube chunks have extra right-side margins that are pushing the videos to the left, which makes them flow outside their parent containers.

Examples:
Extra YouTube chunk margins in a question's feedback section:
Screen Shot 2022-09-21 at 9 44 56 AM

Extra YouTube chunk margins in a question's explanation section:
Screen Shot 2022-09-21 at 9 45 06 AM

It looks like the extra margins on YouTube chunks are an (probably) unintended consequence of the classes being applied to those chunks. Specifically the `.obojobo-draft--chunks--question .solution .solution-container > div > .component[data-obo-component='true']` rule is applying margins on the left and right side.

Should be easy enough to adjust the specificity of this rule to omit YouTube chunks somehow.

FrenjaminBanklin:
**This release has migrations.**

Closes #1272.
Closes #1644.
Closes #1770.
Closes #1788.
Closes #1922.
Closes #1983.
Closes #2012.
Closes #2026.

This release:
- Prevents text styling inside HTML chunks.
- Adds aria-live to the modal component so the screenreader will read any text updates inside, this affects the response dialog after submitting an assessment.
- Adds a response status cache and aria-live property to QuestionResponseStatus to keep track of statuses already read and avoid repeating announcements. Whenever a new status is received, text inside a hidden div is updated and read by the screen reader.
- Prevents questions from forgetting their 'Survey Only' status when changing type between Multiple Choice and Numeric.
- Prevents modules from being saved if there are any `input`s or `select`s with erroneous/invalid values, such as if a numeric question has an invalid string value or an answer of '0' and allows for a percentage margin of error.
- Allows users to specify levels of access to their modules instead and prevents users from accidentally revoking their own access to their modules.
- Fixes positioning of YouTube chunks within question/answer/feedback/explanation boxes to prevent them from rendering outside of their containers.
- Fixes the location of the outline highlight when tabbing between modules in the dashboard.
- Prevents issues inserting events attached to requests forwarded through multiple proxies.
- Adds author-defined module/page/content learning objectives.
- Bumps `moment` version to 1.14.8.
- Bumps `terser` version to 5.14.2.
- Bumps `file-type` version to 16.5.4.
- Bumps `loader-utils` version to 1.4.2.
- Bumps `decode-uri-component` version to 0.2.2.
- Bumps `qs` version to 6.5.3.


FrenjaminBanklin:
This has been running successfully in production in its (more or less) current form.

Committing this to `master` so we can target dev/30 for new work.

PR #2024, opened 2022-08-29
Prevent styling inside HTML chunks 

### LINKED ISSUE #1272: Prevent styling inside the HTML chunk

deundrewilliams:
Carrying over from https://github.com/ucfopen/Obojobo/pull/1920

fixes #1272 

Based on feedback from the linked PR, this disables all of the styling buttons in the toolbar when the cursor is inside an HTML node.


FrenjaminBanklin (approved):
Styling buttons are disabled inside HTML blocks and no styles previously applied inside HTML blocks stick post-update.

PR #2027, opened 2022-09-21
Update numeric question validity checks

### LINKED ISSUE #2012: Numeric questions with an answer of '0' can not have a percentage margin of error.

deundrewilliams:
Resolves #2012 

The initial issue was that the editor allows saving when a margin of error question is created with an answer of 0 and an error type of 'Percent', which would cause a blank screen error when viewing the assessment. This also occurs if any other input for numeric questions is invalid (range invalid, answer can't be parsed to a number, etc.). The first part of these changes adds a validity check to make sure that a margin of error question doesn't have an answer of 0 with an error type of 'Percent'. The other part adds a check before saving that checks the validity of all `input` and `select` fields. If any are invalid, the module won't be saved and an alert will be shown, so this will prevent any type of invalid input from being saved.

### DIFF
--- a/packages/app/obojobo-document-engine/__tests__/oboeditor/components/visual-editor.test.js
+++ b/packages/app/obojobo-document-engine/__tests__/oboeditor/components/visual-editor.test.js
@@ -1451,4 +1451,25 @@ describe('VisualEditor', () => {
 		expect(spy).not.toHaveBeenCalled()
 		expect(instance.state.saveState).toBe('')
 	})
+
+	test('PageEditor component doesnt save if any input field is invalid', () => {
+		const props = { model: { title: '' } }
+		const mockFn = jest.fn()
+		const spy = jest.spyOn(VisualEditor.prototype, 'exportCurrentToJSON')
+
+		const mockReportValidty = jest.fn().mockReturnValue(false)
+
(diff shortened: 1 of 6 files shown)


FrenjaminBanklin (approved):
This is a succinct change that works great. Well done!

I think I'd like to see this embellished a bit - rather than a standard browser alert, maybe using a fancier-looking component that provides specific details about any existing issues and potentially can move focus to the issues for quicker recovery.

For now, this is a great start.

PR #2029, opened 2022-09-26
issue/2026 Fixing Youtube Video Margins

RachelDau:
Added two CSS styles in viewer-component that reverse the applied margin styles to YouTube videos placed in Explanation and Feedback portions of the assessment chuck. This will stop the videos from being off center.

### DIFF
--- a/packages/obonode/obojobo-chunks-multiple-choice-assessment/MCChoice/viewer-component.scss
+++ b/packages/obonode/obojobo-chunks-multiple-choice-assessment/MCChoice/viewer-component.scss
@@ -77,6 +77,10 @@
 				padding-left: 0;
 				padding-right: 0;
 				margin-left: -0.4em;
+
+				.obojobo-draft--chunks--you-tube {
+					margin-left: 0.4em;
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


deundrewilliams (approved):
Tested locally and the changes lgtm

FrenjaminBanklin (changes requested):
This addresses most of the issues, but it looks like the negative left padding is still occurring in feedback for numeric questions. Cover that case as well and this should be good to go.

I get the feeling there's the potential for a solution that doesn't require addressing YouTube chunks specifically, but otherwise this looks fine for now.

FrenjaminBanklin:
I suspect the text nodes have the negative margin to mitigate some kind of baseline padding so they align better along the left side of the container. Unless you find some scenario where the text actually escapes the left boundary, I'd leave it alone for now.

RachelDau:
image
Videos embedded in the question option body are uncentered, but I think this is ok because they line up on the left with the text. Should this also be adjusted?

FrenjaminBanklin:
The right border doesn't extend past the parent container so it's not -too- bad, but it would look better if it was centered properly.

Feel free to take a look, just be careful if the rule that's pushing the video off-center is also factored into the appearance of other things.

RachelDau:
Alright. Thank you. If you get a chance can you check out numeric questions feedback section videos because they appear to be centered on my end. I checked in firefox, chrome, and safari. Let me know if it still there and if you have any tips to reproduce the issue for me to look into.

PR #2031, opened 2022-10-06
Implemented Red Alert Status

Viggie1:
Implemented a red alert status that persists on refresh and logs value into the backend.

Finally finished intro 🎉


FrenjaminBanklin:
Should target the latest dev branch, but aside from that and potentially nitpicking where things should go this looks fine as far as the usual onboarding experience is concerned. Closing.

PR #2032, opened 2022-10-10
Issue/2025 When enter/return key is pressed the input is saved

RachelDau:
Since onFormSubmit function was already implemented to listen and trigger when the enter key is pressed in practice mode, I went there and added code in the preexisting if statement that sends the response to be saved in assessment mode. Now a keyboard user is able to use the enter key to save numeric input responses.

### DIFF
--- a/packages/obonode/obojobo-chunks-question/viewer-component.js
+++ b/packages/obonode/obojobo-chunks-question/viewer-component.js
@@ -32,7 +32,6 @@ export default class Question extends React.Component {
 		this.assessmentComponentRef = React.createRef()
 		this.resultsRef = React.createRef()
 		this.explanationRef = React.createRef()
-		this.questionRef = React.createRef()
 
 		this.onClickBlocker = this.onClickBlocker.bind(this)
 		this.onClickReset = this.onClickReset.bind(this)
@@ -512,7 +511,6 @@ export default class Question extends React.Component {
 				questionIndex={this.props.questionIndex}
 				moduleData={moduleData}
 				resultsRef={this.resultsRef}
-				questionRef={this.questionRef}
 				assessmentComponentRef={this.assessmentComponentRef}
 				type={type}
 				mode={mode}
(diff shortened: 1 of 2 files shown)


FrenjaminBanklin (changes requested):
There are some browser-based issues that are now evident as a result of this change.

The answer is now saved during actual assessments, so that's taken care of. But in Chrome, the input now loses focus, meaning users reliant on a screen reader are required to tab back into the input (and possibly navigate through other tabbables to get there).

In Firefox and Safari this isn't an issue - the input remains focused, but we'll have to do something to address Chrome.

You'll notice in the `componentDidUpdate` of the `Question` viewer component that a switch is checking the state of `this.nextFocus` and shifting focus around the page as necessary. I would suggest creating a new condition here to refocus the answer input upon pressing the 'Enter' key.

RachelDau:
OnEnterPress function was added to delay the refocus of numeric input on enter press in order to to avoid refocusing when a lot of rendering is going on. Unit test added to test that focus is executed at the delay. Now upon enter press the input should retain focus and the answer be saved on all browser types.

packages/obonode/obojobo-chunks-question/viewer-component.js
```diff
@@ -242,6 +248,12 @@ export default class Question extends React.Component {
 		this.setState({ isFlipping: true })
 		setTimeout(() => this.setState({ isFlipping: false }), DURATION_FLIP_TIME_MS)
 	}
+	//delays refocus on input to account for saved text rendering causing a visual loss of focus
+	onEnterPress() {
+		setTimeout(() => {
+			this.assessmentComponentRef.current.inputRef.current.focus()
+		}, 80)
```
FrenjaminBanklin, line 255: This is still kind of a magic number. I'd set it to a constant at the top of the file and use that here instead, like how `DURATION_FLIP_TIME_MS` is used above.

FrenjaminBanklin (changes requested):
One really minor nitpick, see below.

Otherwise this looks like it's good to go.

packages/obonode/obojobo-chunks-question/viewer-component.js
RachelDau: Const added. It is still kinda a magic number though. It was the lowest number that reliably delayed refocus enough to retain focus on all browsers.

FrenjaminBanklin (approved):

PR #2034, opened 2022-10-27
2030 numeric answer validation doesnt detect shorter non numeric strings

Viggie1:
Original issue was to fix the parser so it would not read lowercase letters a-f as hex. A work around was found, but we were concerned with how it would affect Obojobo as a whole. Instead, I implemented a tooltip in the editor that displays to instructors how they can input their answers.

### DIFF
--- a/packages/obonode/obojobo-chunks-numeric-assessment/NumericAnswer/numeric-option.js
+++ b/packages/obonode/obojobo-chunks-numeric-assessment/NumericAnswer/numeric-option.js
@@ -18,6 +18,7 @@ import {
 } from '../entry/numeric-entry-statuses'
 import NumericEntryRange from '../range/numeric-entry-range'
 import isRefRelatedTarget from './is-ref-related-target'
+import NumericInputMoreInfoButton from '../numeric-input-more-info-button'
 
 const RANGE_ERROR_SINGULAR = 'singular'
 const RANGE_ERROR_INVERTED = 'inverted'
@@ -157,6 +158,9 @@ const NumericOption = ({ numericChoice, onHandleInputChange, onHandleSelectChang
 				<div className="is-type-range">
 					<label className="select requirement">
 						Answer Type
+						<div style={{ display: 'inline', position: 'relative', left: 0 + 'em', top: 0 + '%' }}>
+							<NumericInputMoreInfoButton></NumericInputMoreInfoButton>
+						</div>
 						<select
 							className="select-item"
 							name="requirement"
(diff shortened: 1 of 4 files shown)


packages/obonode/obojobo-chunks-numeric-assessment/NumericAnswer/numeric-option.js
```diff
@@ -157,6 +158,9 @@ const NumericOption = ({ numericChoice, onHandleInputChange, onHandleSelectChang
 				<div className="is-type-range">
 					<label className="select requirement">
 						Answer Type
+						<div style={{ display: 'inline', position: 'relative', left: 0 + 'em', top: 0 + '%' }}>
```
FrenjaminBanklin, line 161: Since this same rule is used multiple times throughout the component, either set it as a constant at the top of the file (e.g. `const MORE_INFO_STYLE = {...}`) or set a `className` on the relevant components and define these style rules in an adjacent `.scss` file.

I prefer the second approach.

packages/obonode/obojobo-chunks-numeric-assessment/NumericAnswer/numeric-option.js
```diff
@@ -201,6 +205,9 @@ const NumericOption = ({ numericChoice, onHandleInputChange, onHandleSelectChang
 				<div className="is-type-margin">
 					<label className="select requirement">
 						Answer Type
+						<div style={{ display: 'inline', position: 'relative', left: 0 + 'em', top: 0 + '%' }}>
```
FrenjaminBanklin, line 208: See above comment re: className and adjacent .scss file.

packages/obonode/obojobo-chunks-numeric-assessment/NumericAnswer/numeric-option.js
```diff
@@ -257,6 +264,9 @@ const NumericOption = ({ numericChoice, onHandleInputChange, onHandleSelectChang
 				<div className="is-type-exact">
 					<label className="select requirement">
 						Answer Type
+						<div style={{ display: 'inline', position: 'relative', left: 0 + 'em', top: 0 + '%' }}>
```
FrenjaminBanklin, line 267: See above comment re: className and adjacent .scss file.

FrenjaminBanklin (changes requested):
This looks good, only thing I can see that needs doing is swapping out the inline styles with a stylesheet and a class.

Also, for future reference - **fork this repository** and do all work on an `issue/#-name` branch on **your fork**. It's fine this one time, but in the future make sure that all branches are opened on your own fork and not the official fork.

Viggie1:
This PR has been replaced by #2062

PR #2037, opened 2022-11-14
Issue/1770 Offset focus outline on gems

RachelDau:
I used the `outline-offset` property to move the focus around the gems under `.repository--module-icon` tag out 9px, so that focus is spaced around them properly. This should only effect focus of the gems in the module menu.

### DIFF
--- a/packages/app/obojobo-repository/shared/components/module.scss
+++ b/packages/app/obojobo-repository/shared/components/module.scss
@@ -1,10 +1,12 @@
 @import '../../client/css/defaults';
 
 .repository--module-icon {
+	$padding-module-icon: 0.8em;
+
 	position: relative;
 	font-family: $font-default;
 	width: $dimension-module-icon;
-	padding: 0.8em;
+	padding: $padding-module-icon;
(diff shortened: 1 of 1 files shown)


packages/app/obojobo-repository/shared/components/module.scss
```diff
@@ -23,6 +23,10 @@
 		word-wrap: break-word;
 	}
 
+	:focus {
+		outline-offset: 9px;
```
FrenjaminBanklin, line 27: `9px` seems like it's kind of a magical number in this situation. Ideally it would match some other number which is causing the focused element to be offset from the edges of its parent in the first place.

There's an earlier rule for `.repository--module-icon` that sets `padding` to `0.8em`. Most likely this is the value you'd need to use. Ideally could set this to a variable and use that variable in both places for consistency.

FrenjaminBanklin (changes requested):
One quick fix and this will be ready to go.

FrenjaminBanklin (approved):
Looks good to me.

PR #2044, opened 2022-12-05
issue/2042 - lists render incorrectly in Firefox

RachelDau:
The style causing the behavior was a `display:block;` style on obo-text. At first I tried to overwrite this style, but most ways were ineffective (not a high enough priority to overrride) and when I did override it is also affected all the obo-text on the page changing it all to inline. This didn't appear to change the appearance, but I don't want to affect all text.

My current proposal is to remove the specific line that specifies the display as block at a high level since it the removal doesn't appear to affect anything else (other defined display styles take over) and allows the lists to render correctly as their default. However, this is a bit like removing a fence when you don't entirely know why it is there because I don't know if this style is necessary from some reason I am not aware and didn't visually notice. Feedback and suggestions are welcome.

### DIFF
--- a/packages/app/obojobo-document-engine/src/scss/main.scss
+++ b/packages/app/obojobo-document-engine/src/scss/main.scss
@@ -239,22 +239,4 @@ body {
 			text-align: center;
 		}
 	}
-
-	.obojobo-draft--chunks--list {
-		.obo-text {
-			display: inline;
-
-			&.align-left {
-				text-align: left;
-			}
-
-			&.align-right {
-				text-align: right;
-			}
-
-			&.align-center {
-				text-align: center;
-			}
-		}
-	}
 }
--- a/packages/app/obojobo-document-engine/src/scss/main.scss
+++ b/packages/app/obojobo-document-engine/src/scss/main.scss
@@ -225,6 +225,8 @@ body {
 	}
 
 	.obo-text {
+		display: block;
+
 		&.align-left {
 			text-align: left;
 		}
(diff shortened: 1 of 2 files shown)


FrenjaminBanklin (changes requested):
I agree that removing this rule has potential unforeseen consequences, however I don't think it's impossible to override it.

The actual specificity at play in this rule is `.component[data-obo-component='true'] .obo-text`, so you'd have to supersede that in order to override the style. It should be possible to use `.obojobo-draft--chunks--list` in some capacity to only change styling on `.obo-text` elements within lists.

RachelDau:
Alright, I found a way to override the block style for `.obojobo-draft--chunks--list` in the main.scss file. It appears to fix the Firefox problem and not affect other browsers in a negative way in my testing.

FrenjaminBanklin (changes requested):
`.obojobo-draft--chunks--list` is node-specific, so this rule should be applied in the relevant stylesheet for that obonode rather than here - probably `packages/obonode/obojobo-chunks-list/viewer-component.scss`.

RachelDau:
Alright, this may take a little longer then, because none of my testing for putting it in that file override the other block style. I will see what else I can figure out about how to override this.

RachelDau:
Override inline display style now in the viewer components file. Special thanks to Brandon for helping me out with this.

FrenjaminBanklin (approved):
Looks like this does the trick, and is pretty concise.

PR #2054, opened 2023-01-09
Use Node 16.

### LINKED ISSUE #2013: Upgrade Node to version 18.

FrenjaminBanklin:
Closes #2013.

Bumped Node engine version to 16.19.0. Adjusted Dockerfile base image accordingly. Changed test expectations for 'property of undefined' error message.


RachelDau (dismissed):
I ran this branch on Safari, Chrome, and Firefox and it all looks good. I wasn't sure how to test other than just ensuring all was running and operating as normal.

RachelDau (approved):
LGTM

FrenjaminBanklin:
Closing in favor of #2098.

PR #2055, opened 2023-01-10
Prevent saving drafts or updating permissions for users without appropriate access levels. Explain access levels.

### LINKED ISSUE #2040: Explain access levels. Prevent module actions outside of a user's access level.
See #1983.

#1997 adds the ability to grant any of three different levels of access to a module. The dashboard correctly interprets the current level of access a user has to any given module at the time of the page loading - however, if a user loads the page and their access level is changed before they load the page again, they are still able to interact with the module based on their previous level of access.

For example: User A previously granted 'Full' access to User B for Module X. User B visits their dashboard and can see Module X. User A changes User B's access to 'Partial' or even 'Minimal', but User B has not refreshed their dashboard. Despite now having non-'Full' access to Module X, User B is still able to manage Module X as though they have 'Full' access - including editing, deleting, or changing other users' access levels.

The differences between each access level are also not apparent to the user setting the levels. This could be solved either with some kind of succinct explanatory text in the Share dialog or perhaps a tooltip.

FrenjaminBanklin:
Resolves #2040.

Explanatory text added to Module Permissions window explaining what each access level allows.

Added logic to draft update API endpoint to prevent saving a draft if the user does not have full or partial access to it.

Added logic to permissions update API endpoint to prevent a user without full access to a draft to change other users' access to it.

### DIFF
--- a/packages/app/obojobo-express/__tests__/routes/api/drafts.test.js
+++ b/packages/app/obojobo-express/__tests__/routes/api/drafts.test.js
@@ -410,7 +410,7 @@ describe('api draft route', () => {
 				expect(response.body.value).toHaveProperty('type', 'notAuthorized')
 				expect(response.body.value).toHaveProperty(
 					'message',
-					'Your access level must be "Partial" or higher to retrieve this information.'
+					'In order to edit this module you must have "Partial" or "Full" access.'
 				)
 			})
 	})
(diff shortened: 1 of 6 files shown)


RachelDau:
I tested this on chrome, safari and firefox and it looks like it is working well. I can share modules with users with different access levels and change those and it behaves as described.

RachelDau (approved):

PR #2056, opened 2023-01-12
Adds a filter to the Module Library page.

### LINKED ISSUE #2000: Add a filter to the community collection page.
Currently, all of the modules listed in the community collection are shown in a single huge list, which can make it hard to find something specific. Allow the community library to be filtered like the personal library.

At some point if it becomes possible to have collections inside other collections, it'd be nice to do this in the community library as well.

FrenjaminBanklin:
Resolves #2000.

Module library page now hydrates to enable live filtration.

Module/collection filter functions were moved into their own file to be shared easily between components.

Added a function to the collection backend model to enable looking up all public collections, whenever that becomes necessary.

### DIFF
--- a/packages/app/obojobo-repository/client/css/_defaults.scss
+++ b/packages/app/obojobo-repository/client/css/_defaults.scss
@@ -12,6 +12,7 @@ $color-dangerous-minor: desaturate($color-dangerous, 20%);
 $color-text: #000000;
 $color-text-minor: lighten($color-text, 40%);
 $color-text-subheading: #a20e83;
+$color-text-placeholder: #aaaaaa;
 $color-bg: #ffffff;
 $color-bg-overlay: rgba(200, 200, 200, 0.9);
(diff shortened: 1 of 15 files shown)


RachelDau:
Looks like the filter works well when I tested the branch. I noticed when you click into the filter search on **chrome** a popup of '0' comes up and goes away when you type a character, but I am not sure if it is related to this PR. It only happens with chrome on my computer.
image
Screen Shot 2023-01-19 at 8 12 38 AM

RachelDau:
I think this is some sort of suggested auto fill thing on chrome and not related to the changes, but I am not sure how it works so I figured I would bring it up.

FrenjaminBanklin:
The suggested text is definitely not an Obojobo thing, so I don't think it's a problem as far as this PR is concerned.

RachelDau (approved):

RachelDau (approved):
Looks good to me! Locally it is all working on all three browsers for me.

PR #2064, opened 2023-03-02
Issue/2009 Notifications Added

### LINKED ISSUE #2009: Notifications.
It would be nice to have some kind of in-app notifications for important events - upcoming downtime, release changes, etc.

What I'm envisioning is a way for administrators to create announcements and set a date for when those announcements will be displayed. Potentially we could also determine who sees these announcements - only authors, only learners, or both. Whenever a user accesses Obojobo, the system could check if their last login is earlier than the announcement window and, if so, display the announcement to them. If multiple announcements are relevant it could stack them.

This will likely require a new database table. At the moment there are no admin-only interfaces in Obojobo. While it would be nice to have one for this purpose, it may make the most sense to stick with manually plugging information into the database.

RachelDau:
For[ issue 2009](https://github.com/ucfopen/Obojobo/issues/2009) adds notification to obojobo. Right now these display in the Nav section of a module. You must physically type the information into the database `notifications`. It then uses the last login date and compares it to the notification created_at time to determine if the notification should be displayed and which one should be displayed.

Notification will be taken directly from the database, written to the cookie, pulled out of the cookie on the front-end side, and rendered from the dashboard repository nav file. If notifications are present (typed into Database) upon page refresh a red circle indicator will appear by the username when on a dashboard page. When you click the indicator or the username area it will toggle a notifications drop down. From here you can see and exit notifications which removes them from the cookie. If you input new notification into the database, the previous ones will disappear so they must be done together if you want them to appear together. You cannot use certain symbols such as `; `in the cookie title or text and it is processed as a string and this symbol will mess up how the code processes the notifications from the cookie.


packages/app/obojobo-express/server/routes/api/visits.js
```diff
@@ -91,6 +99,26 @@ router
 					visitId: req.currentVisit.id
 				})
 			})
+			.then(() => viewerUsersState.getLastLogin(userId))
+			.then(result => {
+				if (result) lastLogin = result.last_login
+			})
```
FrenjaminBanklin, line 105: Ideally req.currentUser would just have a lastLogin property that you could check, rather than creating a whole new function to get that information separately.

packages/app/obojobo-express/server/routes/api/visits.js
```diff
@@ -91,6 +99,26 @@ router
...
+				if (result) isNotificationEnabled = result.status
+			})
+			.then(() => viewerNotificationState.getTitle(id))
+			.then(result => {
+				if (result) notificationTitle = result.title
+			})
+			.then(() => viewerNotificationState.getText(id))
+			.then(result => {
+				if (result) notificationText = result.text
+			})
```
FrenjaminBanklin, line 121: Try to provide all valid notifications to the frontend in a single array rather than going through multiple steps to grab individual pieces of a single notification like this.

packages/app/obojobo-express/server/routes/api/visits.js
```diff
@@ -111,7 +139,10 @@ router
 					isPreviewing: req.currentVisit.is_preview,
 					lti,
 					viewState,
-					extensions: visitStartExtensions
+					extensions: visitStartExtensions,
+					isNotificationEnabled,
+					notificationTitle,
+					notificationText
```
FrenjaminBanklin, line 145: See above comment: rather than informing the frontend of a single notification with three new properties, try to provide all valid notifications as objects in a single array.

packages/app/obojobo-express/server/viewer/viewer_notification_state.js
```diff
@@ -0,0 +1,57 @@
...
+		`
+		SELECT text FROM notification_status
+		WHERE
+			id = $[id]
+	`,
+		{
+			id
+		}
+	)
+}
```
FrenjaminBanklin, line 38: Don't create multiple functions for each individual field, just have a single `getNotifications(ids)` function to grab multiple notifications - pass it an array (could even be an array containing a single number) and handle results accordingly.

packages/app/obojobo-express/server/viewer/viewer_notification_state.js
```diff
@@ -0,0 +1,57 @@
...
+		`
+			SELECT id FROM notification_status
+			WHERE created_at >= $[lastLoginDate]
+			LIMIT 1
+		`,
+		{
+			lastLoginDate
+		}
+	)
+}
```
FrenjaminBanklin, line 50: Several things here:

- Change the name from `getId` to something more descriptive like `getRecentNotifications`.
- Generalize `lastLoginDate` to `targetDate` or maybe even just `date`.
- Use `manyOrNone` instead of `oneOrNone` and remove the `LIMIT` so this function grabs all notifications after a given date.
- Add an `ORDER BY created_at ASC` so notifications always appear in a predictable order.

packages/app/obojobo-express/server/viewer/viewer_users_state.js
```diff
@@ -0,0 +1,41 @@
...
+		{
+			userId
+		}
+	)
+}
+
+module.exports = {
+	//setLastLogin,
+	getLastLogin
+}
```
FrenjaminBanklin, line 41: See if it's possible to attach the last login date to `req.currentUser` instead of creating a separate function like this to look it up separately.

packages/app/obojobo-express/server/migrations/20221214210826-modify-users-add-last-login-column.js
```diff
@@ -0,0 +1,31 @@
...
+exports.up = function(db) {
+	return db.addColumn('users', 'last_login', {
+		type: 'timestamp WITH TIME ZONE',
+		notNull: true,
+		defaultValue: new String('now()')
+	})
+}
+
+exports.down = function(db) {
+	return null
```
FrenjaminBanklin, line 26: The `down` of a migration is the opposite of the `up`. In this case, the `down` of this migration should remove the `last_login` column.

FrenjaminBanklin (commented):
Initial thoughts:

You might need to describe the new `notification_status` table a bit. It's not clear what the `user_id`, `draft_id` or `status` fields are for. It doesn't look like `draft_id` is made anywhere.

Ideally all available notifications should be objects in an array and handled programmatically instead of explicitly defining `notificationText` and `notificationTitle` to handle only one notification at a time.

If the idea for notifications is that we aren't tracking which ones have been seen by which users, then the three fields mentioned above shouldn't be necessary. In order to preserve notifications between page loads, look into adding them to a cookie (see [here](https://github.com/ucfopen/Obojobo/blob/f270fb5465df8c4457a8dd003e6e2f78dfebeaec/packages/app/obojobo-repository/server/routes/dashboard.js#L31-L33) for an example of using cookies) and changing the utilities to adjust what's saved in the cookies.

More specific comments per file below.

RachelDau:
While previously mentioned issues have been addressed current known issues now include:

- last_login is not updating in the user table 
- notifications are no re-rendering upon exit button press
- removing the notifications from the cookie has not successfully been implemented
- unit tests for visits.js changes are unwritten... and there are jest failures from tests in that file
- a CSS cleanup and style improvement is needed for the notifications

RachelDau:
merged with latest dev branch to solidify new node version , no notifications changes were included in the merge.

RachelDau:
Previous Issues have been addressed current issues:

- cannot exit last notification in a set
- it would be nice to auto-scroll to the top notification when the dashboard is opened
- improvement to notification design/ obo standards (feedback welcome)

RachelDau:
Notifications now should all exit when clicked and are removed from cookie upon refresh. 

To test Notifications, migrate up the migration files to create a local notifications table. Input the information for each notification directly in the table (id, text, and title). Notifications from this table will display upon dashboard reload. They will persist until a user exits them, logs in again, or they will also disappear if more notifications are added (only the new ones will be displayed). You must scroll up on the dashboard page to see them because I do not have autoscroll up to their location working currently.

FrenjaminBanklin:
When possible, update the PR's description so that it accurately reflects the current state of work on the PR.

packages/app/obojobo-repository/shared/components/notification.scss
```diff
@@ -0,0 +1,62 @@
+@import '../../../obojobo-document-engine/src/scss/_includes.scss';
```
FrenjaminBanklin, line 1: You should be able to replace this line with `@import '~styles/includes';`.

packages/app/obojobo-repository/shared/components/notification.scss
```diff
@@ -0,0 +1,62 @@
...
+$focus-pink: #fbdae6;
+$white: white;
+
+h1 {
+	margin: 0;
+}
+
+p {
+	display: block;
+}
```
FrenjaminBanklin, line 13: Try to nest these rules under whichever class rule below is/are relevant. There's too much of a chance that having them free-floating like this could cause conflicts elsewhere on the page.

packages/app/obojobo-express/server/migrations/20221214210826-modify-users-add-last-login-column.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
FrenjaminBanklin, line 1: I would probably regenerate these migrations so their filenames indicate a more recent date, to make sure they're run in the correct order.

packages/app/obojobo-express/server/migrations/20230301155626-create-notification.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
```
FrenjaminBanklin, line 1: I would probably regenerate these migrations so their filenames indicate a more recent date, to make sure they're run in the correct order.

packages/app/obojobo-repository/shared/components/layouts/default.jsx
```diff
@@ -3,6 +3,8 @@ require('./default.scss')
 const React = require('react')
 const Footer = require('./footer')
 const Dedication = require('./dedication')
+const Notification = require('../notification.jsx').default
```
FrenjaminBanklin, line 6: After moving the `Notification` component from this file into the `Dashboard` component, this import will be unnecessary.

See suggested changes for the `notification.jsx` file, which would allow you to `require` this the usual way instead of specifying `.default`.

packages/app/obojobo-repository/shared/components/layouts/default.jsx
```diff
@@ -53,6 +55,9 @@ const LayoutDefault = props => (
 		</head>
 		<body className={props.className}>
 			<div className="layout--wrapper">
+				<span id="repository-notifications" data-react-props={'{}'}>
+					<Notification></Notification>
+				</span>
```
FrenjaminBanklin, line 60: See above comment re: moving this into the `Dashboard` component.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import { useState, useEffect } from 'react'
```
FrenjaminBanklin, line 2: This could be shortened as `const React, { useEffect, useState} = require('react')`.

You could also just `const React = require('react')` and use `React.useState` and `React.useEffect` in the code instead, if you're only going to do so once or twice.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,67 @@
...
+import './notification.scss'
```
FrenjaminBanklin, line 3: Likewise, this would be `require('./notification.scss')`.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,67 @@
...
+					renderNotification(key, notifications.text, notifications.title)
+				)}
+			</div>
+		)
+	} else {
+		return null
+	}
+}
+
+export default Notification
```
FrenjaminBanklin, line 67: This can be changed to `module.exports = Notification`, which should allow it to be `require`'d or `import`'d normally.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,67 @@
...
+				<p>{text}</p>
+			</div>
+		)
+	}
+
+	//rewrite to cookie to remove notifications that have been hidden
+	if (hiddenNotifications && hiddenNotifications.length >= 1) {
+		const jsonNotifications = JSON.stringify(notifications)
+		const cookieString = `${encodeURIComponent(jsonNotifications)};`
+		document.cookie = 'notifications=' + cookieString
```
FrenjaminBanklin, line 52: These three lines can be moved into the `onClickExitNotification` function and the `hiddenNotifications` state variable can be eliminated entirely.

packages/app/obojobo-repository/shared/components/pages/page-dashboard-client.jsx
```diff
@@ -1,8 +1,10 @@
...
+import { hydrateEl, hydrateElWithoutStore } from '../../react-utils'
 import Dashboard from '../dashboard-hoc'
 import DashboardReducer from '../../reducers/dashboard-reducer'
+import Notification from '../notification.jsx'
 
 // include LayoutDefault so client side react has a copy of it
 import LayoutDefault from '../layouts/default' //eslint-disable-line no-unused-vars
 
 hydrateEl(Dashboard, DashboardReducer, '#react-hydrate-root')
+hydrateElWithoutStore(Notification, '#repository-notifications')
```
FrenjaminBanklin, line 10: Moving the notification component into the Dashboard and rendering it exclusively on the client side will remove the need for server-side rendering and hydration, so this line (and those pointed out above) can go.

packages/app/obojobo-repository/shared/components/pages/page-dashboard-client.jsx
```diff
@@ -1,8 +1,10 @@
-import { hydrateEl } from '../../react-utils'
+import { hydrateEl, hydrateElWithoutStore } from '../../react-utils'
```
FrenjaminBanklin, line 1: Since the notification component is no longer being rendered/hydrated, `hydrateElWithoutStore` is no longer a necessary import here.

packages/app/obojobo-repository/shared/components/pages/page-dashboard-client.jsx
```diff
@@ -1,8 +1,10 @@
...
 import Dashboard from '../dashboard-hoc'
 import DashboardReducer from '../../reducers/dashboard-reducer'
+import Notification from '../notification.jsx'
```
FrenjaminBanklin, line 4: Moving the notification component into the Dashboard and rendering it exclusively on the client side will remove the need for server-side rendering and hydration, so this import can go.

FrenjaminBanklin (changes requested):
Now that I'm seeing this in practice, I can see a few things I'd like to change. Aside from relatively minor changes in a few files, the big things I could see addressing are as follows:

First:
Since we've more or less decided to focus on keeping notifications in the dashboard for now, and the actual rendering of those notifications only occurs on the frontend, I think the extra mechanics for adding it to the default layout and hydrating it are unnecessary. More context in some file comments below.

In `dashboard.jsx`, add a `const Notification = require('./notification')` to the import stack at the top and add `<Notification />` as the first child under `#dashboard-root`, just before the `RepositoryNav` component.

Second:
Getting notifications rendering on the page is great progress. Not making them visible on the screen right away is a bit of a problem, though - and the current format of them is a bit intrusive even if you can get that figured out.

I'm wondering how feasible it would be to start with the notifications hidden behind some kind of clickable indicator. Maybe something like this:
image

Or (probably better) like this:
image

Clicking this indicator would show the notifications in some way. My first thought would be a vertically scrolling list that appears  between the nav area and the 'Dashboard' banner (or placed absolutely to overlap the 'Dashboard' banner). A modal dialog window would probably be cleaner and less intrusive, though.

packages/app/obojobo-repository/shared/components/notification.jsx
RachelDau: ES Modules may not assign module.exports or exports.*, Use ESM export syntax, instead: ../obojobo-repository/shared/components/notification.jsx

I switched to writing this ESM syntax because of this error I got when trying to use the component in the repository-nav file. I still get this error when I change it and at this moment I am still not sure why, but looking into it

RachelDau: I figured out it was my import React from 'react' that needed to be switched to a const React = require('react') right after this haha

packages/app/obojobo-repository/shared/components/notification.jsx
RachelDau: The issue I encountered here is that if they are in the onClickExitNotification function then the notifications exit and remove from the page, however, they are re-rendered on page refresh and so the exit does not persist after refresh.  This was because on refresh the cookie is retaken and the state variables redefined. Any change to the state variable notifications is undone in refresh. However, if I exited two notifications the first one would be removed and only the second one would persist. So I created a list of notifications to be hidden called hidden notifications and added the if statement to update the cookie after the exit button is pressed, but before refreshing the page so that refreshing wouldn't lose the last exited change. This was a work-around so I will look into it more to see if there is a better solution.

RachelDau:
New Updates: When notifications are present a circle indicator will appear next to the username and when clicked a dropdown of the notifications reveals itself. When clicked again it will disappear. I also edited css styles formatting and fixed my imports. Will re-generate migration files and push after addressing the feedback on the notification component exiting issue.

RachelDau:
Brandon's feedback on notifications from the last iteration has been addressed. In order to test locally, you must migrate the db up to create the new notification table and last login in the users table. `yarn db:migrateup` You then must type directly in the database notification table the content for the notification (id, title, and text). Let me know if you need help.

Notification will be taken directly from the database, written to the cookie, pulled out of the cookie on the front-end side, and rendered from the dashboard repository nav file. If notifications are present (typed into Database) upon page refresh a red circle indicator will appear by the username when on a dashboard page. When you click the indicator or the username area it will toggle a notifications drop down. From here you can see and exit notifications which removes them from the cookie. If you input new notification into the database, the previous ones will disappear so they must be done together if you want them to appear together. You cannot use certain symbols such as ; in the cookie title or text and it is processed as a string and this symbol will mess up how the code processes the notifications from the cookie.

packages/app/obojobo-repository/shared/components/repository-nav.scss
```diff
@@ -1,5 +1,7 @@
 @import '../../client/css/defaults';
 
+$poppin-pink: #af1b5c;
```
FrenjaminBanklin, line 3: Since this color is used in two places, remove the declarations from the individual stylesheets and put them in a single location used by everything that requires the color.

Also name more descriptively, e.g. `$color-notification`.

packages/app/obojobo-repository/shared/components/repository-nav.jsx
```diff
@@ -1,12 +1,15 @@
...
-const { useState } = require('react')
 const Avatar = require('./avatar')
+const Notification = require('./notification')
 
 const RepositoryNav = props => {
 	let timeOutId
-	const [isMenuOpen, setMenuOpen] = useState(false)
+	const [isMenuOpen, setMenuOpen] = React.useState(false)
+	const [isNotificationsOpen, setNotificationsOpen] = React.useState(false) // New state for notification
+	const [notificationsExist, setNotificationsExist] = React.useState(false)
```
FrenjaminBanklin, line 11: There doesn't appear to be anything in the file using `setNotificationsExist`. It also looks like the notification indicator remains on the screen even if all notifications are dismissed.

Consider passing `setNotificationsExist` into the `Notification` component and adding some logic to that component to call it with `false` if all notifications are dismissed.

packages/app/obojobo-repository/shared/components/repository-nav.jsx
```diff
@@ -49,7 +77,12 @@ const RepositoryNav = props => {
 				{props.userId !== 0 ? (
 					<div className="repository--nav--current-user">
 						<button onClick={onToggleMenu}>
-							<div className="repository--nav--current-user--name">{props.displayName}</div>
+							<div onClick={onNotifications} className="repository--nav--current-user--name">
+								{props.displayName}
+
+								{notificationsExist && <div className="notification-indicator"></div>}
+							</div>
+
```
FrenjaminBanklin, line 85: I see what you're going for, but I think rather than make the entire username clickable all the time it would be better if the notification indicator was more prevalent, and the notification indicator was listening for clicks rather than the entire username/avatar area. You'll notice how the addition of this `onClick` has altered numerous unrelated snapshot files.

It may be better after all to go with the second option, where notifications are indicated with a clickable '# New Messages' element that appears under the username instead of inline.

packages/app/obojobo-repository/shared/components/notification.scss
```diff
@@ -0,0 +1,57 @@
+@import '~styles/includes';
+
+$poppin-pink: #af1b5c;
```
FrenjaminBanklin, line 3: Since this color is used in two places, remove the declarations from the individual stylesheets and put them in a single location used by everything that requires the color.

Also name more descriptively, e.g. `$color-notification`.

packages/app/obojobo-repository/shared/components/notification.scss
```diff
@@ -0,0 +1,57 @@
...
+$focus-pink: #fbdae6;
```
FrenjaminBanklin, line 4: Same as above - if this color is used in more than one place, move it to a location accessible in all the places it is used.

Also name it more descriptively, e.g. `$color-notification-focus`.

packages/app/obojobo-repository/shared/components/repository-nav.scss
```diff
@@ -119,6 +121,19 @@
 		width: 2.9em;
 		margin: 0;
 	}
+
+	.notification-indicator {
```
FrenjaminBanklin, line 125: Wherever this indicator ends up on the page, make sure there's also a `cursor: pointer;` in there so it's obvious that it can be clicked.

packages/app/obojobo-repository/shared/components/notification.scss
```diff
@@ -0,0 +1,57 @@
...
+		display: block;
+	}
+
+	.notification-header {
+		display: flex;
+		flex-direction: row;
+		justify-content: space-between;
+		align-items: center;
+
+		.notification-exit-button {
```
FrenjaminBanklin, line 29: Make sure there's a `cursor: pointer;` rule in here so it's obvious that this button is clickable.

FrenjaminBanklin (changes requested):
I think this is moving in the right direction.

More specific comments are below, but summarized:
- See if it's possible to move the notifications indicator and change it from a pink dot to a '# messages' string, make it more obvious that it's clickable either way, and give the indicator element the click handler instead of the parent element so unrelated snapshots are not polluted.
  - Also adjust styles on the 'X' button in each notification so that it's obvious those are clickable.
- The indicator for notifications existing persists after all notifications are dismissed, until the page is refreshed.

I didn't leave a specific comment, but I think the notifications still take up a lot of room vertically. I think the font sizes in both the title and body text for notifications can stand to be shrunk a bit.

RachelDau:
Taking into account all previous feedback, notifications are now in a pop-up with some style changes. Some CSS styles were reorganized and jest tested added for the new notification indicator. It now tracks the number of notifications present in a callback function from the notification component to repository-nav. When exiting the indicator updates immediately. 

As usually, after migrating the db up to create the notification table type directly into it to add notifications with a id, title, and text.

Any and all feedback is appreciated :)

packages/app/obojobo-repository/shared/components/repository-nav.jsx
```diff
@@ -49,7 +86,16 @@ const RepositoryNav = props => {
 				{props.userId !== 0 ? (
 					<div className="repository--nav--current-user">
 						<button onClick={onToggleMenu}>
-							<div className="repository--nav--current-user--name">{props.displayName}</div>
+							<div className="repository--nav--current-user--name">
```
FrenjaminBanklin, line 89: This doesn't appear to be reachable in the tab order. I'm not sure how the rest of the page is handling tab order, but the notifications indicator doesn't appear to be part of it.

FrenjaminBanklin (changes requested):
This is looking pretty good.

The main things I can see that still need addressing are tab reachability and click handlers. Currently the notifications indicator is not tab-reachable, so it can only be activated by clicking it directly. When the notifications are visible, only the 'close' buttons are tab-reachable so there's currently no way of screen readers reaching the text content.

It looks like clicking anywhere except one of the notification's 'X' buttons dismisses the notifications popup entirely. Even clicking inside one of the notifications.

RachelDau:
Updated the screen for when all notifications are exited (style update)

RachelDau:
These are ready for review

In order to test locally, you must migrate the db up to create the new notification table and last login in the users table. yarn db:migrateup You then must type directly in the database notification table the content for the notification (id, title, and text). Let me know if you need help.

Notification will be taken directly from the database, written to the cookie, pulled out of the cookie on the front-end side, and rendered from the dashboard repository nav file. If notifications are present (typed into Database) upon page refresh a red circle indicator will appear by the username when on a dashboard page. 

When you click the indicator or the username area it will open a notifications pop-up. From here you can see and exit notifications which removes them from the cookie. If you input new notification into the database, the previous ones will disappear so they must be done together if you want them to appear together. (they will also disappear on re-login) 

You cannot use certain symbols such as ; in the cookie title or text and it is processed as a string and this symbol will mess up how the code processes the notifications from the cookie. Also if the title is one word thats more than 37 characters in length that may be an issue.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,68 @@
...
+	}, [notifications])
+
+	const renderNotification = (key, text, title) => {
+		return (
+			<div className={`notification-banner`} key={key}>
+				<div className="notification-header">
+					<h1 tabindex="0">{title}</h1>
```
FrenjaminBanklin, line 41: It seems to be functionally fine, but it's worth noting that when using React 'tabindex' properties need to be written in camel-case as 'tabIndex' to be considered valid.

packages/app/obojobo-repository/shared/components/notification.jsx
```diff
@@ -0,0 +1,68 @@
...
+				if (parts[0] === 'notifications') {
+					parsedValue = JSON.parse(parts[1])
+				}
+			})
+
+			const parsedNotifications = parsedValue
+			setNotifications(parsedNotifications)
+		} else {
+			// there is nothing to render
+		}
```
FrenjaminBanklin, line 23: You can omit this `else` entirely.

FrenjaminBanklin (changes requested):
Sorry this has taken me so long to review.

This has shaped up really well, the only big thing I can see that needs doing now is controlling tab flow. You'll notice that other modal popups (the Module Options window, the user search sub-window in the Share dialog, etc.) only allow the tab key to reach elements that are within the currently open modal dialog window.

I'm not sure exactly how this is being accomplished, but I have a feeling it's a result of using the `ReactModal` component (see [how it's used in the Dashboard component](https://github.com/ucfopen/Obojobo/blob/e8050b0436b09722e7b2d35a689c230e6f484378/packages/app/obojobo-repository/shared/components/dashboard.jsx#L209-L217)).

If you can figure out how to get the notifications working within ReactModal, you may not have to do as much tab index management yourself.

Also, I feel like we could potentially sanitize cookie string characters such as `;`, but avoiding them should be easy enough. Addressing reserved/unsafe characters can be a future issue.

RachelDau:
Alright I added React modal so that the notifications window is tab accessible and you can't tab back to the main page when accessing. The text and title seem to still need the tabIndex on them to be tab accessible. One thing to note is now clicking outside of the notification window will also close it as this is a sort of default when using modal. Adding modal caused a lot of issues with other tests and so a variety of files that use the repository-nav component and test files for those files needed appropriate mocks of react modal added in order to continue functioning.  A few Dashboard tests needed a slight rewrite because they hinged on their being one modal and now there are two with notifications so the tests were modified to ensure that is reflected.

packages/app/obojobo-express/server/migrations/20240130162639-modify-users-add-last-login.js
```diff
@@ -0,0 +1,31 @@
...
+exports.up = function(db) {
+	return db.addColumn('users', 'last_login', {
+		type: 'timestamp WITH TIME ZONE',
+		notNull: true,
+		defaultValue: new String('now()')
+	})
+}
+
+exports.down = function(db) {
+	return db.removeColumn('last_login')
```
FrenjaminBanklin: `db.removeColumn` takes two arguments, the first being the table name.

This should read `return db.removeColumn('users', 'last_login')`.

packages/app/obojobo-repository/shared/components/notification.test.js
```diff
@@ -0,0 +1,156 @@
...
+		document.cookie =
+			'notifications=' +
+			JSON.stringify([{ key: 1, text: 'Test Notification', title: 'Test Title' }])
+
+		const component = create(<Notification onDataFromNotification={onDataFromNotification} />)
+		const tree = component.toJSON()
+
+		expect(tree).toMatchSnapshot()
+		expect(document.cookie).toBe(
+			'notifications=%5B%7B%22key%22%3A1%2C%22text%22%3A%22Test%20Notification%22%2C%22title%22%3A%22Test%20Title%22%7D%5D;'
```
FrenjaminBanklin: This seems a bit magical, you should be able to store the important value in a variable and use `encodeURIComponent` to check it:
```javascript
const notificationValue = [{ key: 1, text: 'Test Notification', title: 'Test Title' }]

document.cookie = `notifications=${JSON.stringify(notificationValue)}`

...

expect(document.cookie.toBe(`notifications=${encodeURIComponent(notificationValue)}`)
```

packages/app/obojobo-repository/shared/components/notification.test.js
```diff
@@ -0,0 +1,156 @@
...
+		if (exitButtons[key]) {
+			act(() => {
+				exitButtons[key].props.onClick()
+			})
+
+			// Update the tree after state changes
+			tree = component.toJSON()
+			expect(tree).toMatchSnapshot()
+			expect(document.cookie).toBe(
+				'notifications=%5B%7B%22key%22%3A2%2C%22text%22%3A%22Notification2%22%2C%22title%22%3A%22Title2%22%7D%5D;'
```
FrenjaminBanklin: See above comment re: variables and `encodeURIComponent`.

packages/app/obojobo-repository/shared/components/repository-nav.test.js
```diff
@@ -169,4 +203,128 @@ describe('RepositoryNav', () => {
...
+		const reusableComponent = <RepositoryNav {...navProps} />
+		let component
+		act(() => {
+			component = create(reusableComponent)
+		})
+		const tree = component.toJSON()
+		expect(tree).toMatchSnapshot()
+
+		if (document && document.cookie) {
+			//don't get here
```
FrenjaminBanklin: We could probably have some `expect`s above this to check that `document` has a value but `document.cookie` does not, rather than this if/else.

packages/app/obojobo-repository/shared/components/repository-nav.test.js
```diff
@@ -169,4 +203,128 @@ describe('RepositoryNav', () => {
...
+		const tree = component.toJSON()
+		expect(tree).toMatchSnapshot()
+
+		if (document && document.cookie) {
+			const cookiePropsRaw = decodeURIComponent(document.cookie).split(';')
+
+			cookiePropsRaw.forEach(c => {
+				const parts = c.trim().split('=')
+
+				if (parts[0] === 'notifications') {
```
FrenjaminBanklin: See previous comment re: using `expect` to fail early instead of relying on an if/else to check for success.

FrenjaminBanklin (changes requested):
This is nearly ready to go - just a few very minor adjustments to unit tests and one correction to one of the new migrations.

I think the preferred tab order per notification would go title -> message -> close button, but that would require adjusting the order in which the elements are rendered and probably corresponding changes to the styles to make things show up properly. It can wait to be a follow-up issue, I think.

packages/app/obojobo-repository/shared/components/notification.test.js
RachelDau: `expect(document.cookie).toBe(`notifications=${encodeURIComponent(JSON.stringify(notificationValue))};`)`
good idea: had to stringify before encoding or it didn't catch the specific title/text

RachelDau:
I tested migrations up and down and it all appears to be good now. I also regenerated migration files so they are new. Unit test changes were made to reflect using expect and not if else statements.

packages/app/obojobo-repository/shared/components/notification.test.js
```diff
@@ -0,0 +1,151 @@
...
+		})
+
+		let tree = component.toJSON()
+		expect(tree).toMatchSnapshot()
+
+		// Simulate click on exit button for the first notification
+		const key = 0
+		const exitButtons = component.root.findAllByProps({ className: 'notification-exit-button' })
+
+		if (exitButtons[key]) {
```
FrenjaminBanklin: Sorry if I didn't comment on or notice this before. Feels like I keep sending you back for different things every time.

Is there any scenario where this test should run successfully and `exitButtons[key]` would evaluate falsy, or if `exitButtons[key]` is falsy does that mean something went wrong?

Based on the `findAllByProps` line above I would expect a case where `exitButtons[key]` not resolving to something truthy to indicate that there was a problem. In a case like this you could use `https://jestjs.io/docs/expect#tobetruthy`, or forego the `key` variable altogether and `expect(exitButtons.length).toBe(0)` or even `expect(exitButtons.length > 0).toBe(true)`.

packages/app/obojobo-repository/shared/components/notification.test.js
```diff
@@ -0,0 +1,151 @@
...
+		const exitButtons = component.root.findAllByProps({ className: 'notification-exit-button' })
+
+		act(() => {
+			exitButtons[key].props.onClick()
+		})
+
+		tree = component.toJSON()
+		expect(tree).toMatchSnapshot()
+
+		if (elementToExit) {
```
FrenjaminBanklin: This is another case where you should use `expect` rather than `if`. Ideally you should populate `elementToExit` and then have an immediate `expect` to make sure it's what you expect it to be, at which point this `if` would be unnecessary because you've already made sure it exists.

packages/app/obojobo-repository/shared/components/notification.test.js
```diff
@@ -0,0 +1,151 @@
...
+		const onDataFromNotification = jest.fn()
+		const reusableComponent = <Notification onDataFromNotification={onDataFromNotification} />
+		let component
+		act(() => {
+			component = create(reusableComponent)
+		})
+		const tree = component.toJSON()
+		expect(tree).toMatchSnapshot()
+
+		if (document && document.cookie) {
```
FrenjaminBanklin: Since this test is only being run one time and should only be testing one thing, there should probably not be an `if...else` here to determine which expectations are run. In this case you could replace this `if...else` with two expectations:
`expect(document).toBeTruthy()` and `expect(document.cookie).toBeTruthy()`. At that point you can reasonably assume that only the `if` block would ever run, so you wouldn't need the `else` block or the `if...else` at all.

FrenjaminBanklin (changes requested):
I might not have noticed a few things in my last pass, but with a few more minor revisions to the unit tests I think this will be all set.

packages/app/obojobo-repository/shared/components/notification.test.js
RachelDau: oops i totally missed this but rewrote the very similar tests yesterday. Fixing now

packages/app/obojobo-repository/shared/components/notification.test.js
RachelDau: yes, removing the if worked perfectly. Seems it was redundant to check if it existed and then expect it to exist and be undefined in value

RachelDau:
Alright, I removed the if statements from testing as they weren't needed and added more expects to ensure things are tested. When testing make sure to test how it handles no notifications and an empty table. It was working when I tested it I just have a weird fear about that breaking everything.

RachelDau:
image
Something I noticed is because I check that the date is changed to today using a function every so once in awhile the test is off by a second because of the time it takes to set last login and then compare to todays date that a second has passed by the time it is called. Repro rate seems to be every 1/10 tests or so.

FrenjaminBanklin:
Hmm. The values of `today` being off due to execution time is probably something that needs to be resolved now.

Quick Googling brings up [this SO article](https://stackoverflow.com/questions/28504545/how-to-mock-a-constructor-like-new-date) discussing how to mock the constructor for `Date`, which is probably what we'd have to do here to make sure results are consistent.

RachelDau:
My latest commit sets a mock timer to a specific date and using that set date to reference and test, eliminating any dependency on the execution time. Thanks for the link it helped!

FrenjaminBanklin (approved):
Tests are all passing, and I think the code is looking pretty good. Anything I can think to change later will be a separate issue.

PR #2068, opened 2023-03-15
Allow Obonodes to define node types not allowed as children.

### LINKED ISSUE #2066: Prevent certain node types from being nested within certain other node types.
There are situations where it's desirable to allow nodes to be nestable, such as when a question bank has multiple nested question banks that can be selected randomly to add variety to an assessment (worth noting this is mostly driven by buttons inside the question bank that specifically add another question bank). However this isn't always desirable - for example an answer node can have a question bank as a child, which can then have a question, which can then have an answer, which can then have another question bank, and so on. Eventually the only thing stopping more nesting is certain items on screen being scaled down to the point of no longer being usable.

Ideally there would be an easy way for nodes which allow nesting to determine which nodes can be nested in them. Possibilities off the top of my head:

1. Define per component what obonode types are acceptable as children (e.g. questions).
2. Add a property to every obonode that identifies that node type as 'nestable' - then for any component that can have user-defined children only allow 'nestable' components to be children.

I'm open to ideas, since I have no idea what the effort required for either of the above would be, or where there might be additional pitfalls. I'm thinking option 1 allows for more nuance than option 2, but option 2 would probably require less effort overall.

FrenjaminBanklin:
Closes #2066.

Obonode editor registrations can now optionally define a `disallowedChildren` property to indicate node types which can not be children of that node. For example:

```
disallowedChildren: ['ObojoboDraft.Chunks.Question', 'ObojoboDraft.Chunks.QuestionBank'],
```
will automatically omit the Question and QuestionBank options from those provided to the insert menus in the visual editor.


packages/app/obojobo-document-engine/src/scripts/oboeditor/components/node/components/insert-menu.js
```diff
@@ -104,6 +104,9 @@ class InsertMenu extends React.PureComponent {
 
 	renderItem(item, index) {
 		if (item.separator) {
+			// don't display two separators in a row - if the last item was a separator, ignore this one
+			if (index > 0 && this.props.dropOptions[index - 1].separator) return
```
zachberry, line 108: Nice

zachberry (commented):
Insane nitpick time, but what about `disallowedChildren` instead of `canNotContain`?

This works well, the only issue I'm seeing is that the Insert Menu isn't impacted. Maybe menu items could be disabled if they're not allowed to be inserted?

FrenjaminBanklin:
Converting this to a draft while I try to figure out how to also disable options in the Insert menu based on the new settings.

FrenjaminBanklin:
The error pictured has always been there. It'll always pop up once when the visual editor is initialized.

As for getting a ton of errors trying to do anything else - does this happen for all modules, or only modules you had before checking out the new branch? Did you remember to run `yarn` again after checking out the new branch?

RachelDau:
Alright. I went back and ran yarn in the root directory after checking out. This is happening with modules I had previously made and new modules I make. When I click the plus button to add a chunk the screen goes white, but when I refresh everything returns to normal.

FrenjaminBanklin:
It turns out automatically adjusting the 'Insert' menu in the toolbar doesn't work after all.

It looks like that menu is set to only update when the cursor position changes. When moving between existing nodes this works just fine, but when adding new nodes it gets into a weird state. It's both able to find the new node, since the cursor has moved to within that new node, and not able to find that same new node's parent. It's like the node exists and simultaneously doesn't exist.

I'm really struggling to figure out what's going on here and how to fix it intelligently. Currently the process for building the insert menu is memoized so that it only runs when the cursor position changes. Technically the cursor position changes when a new node is added to the document, despite that new node not being 'there' enough that its parent can be determined based on its location. And then when the node is fully rendered (and presumably its parent could be found because it is now actually on the page), the cursor position has technically not changed so the insert menu is not redrawn.  

I could potentially pull this function out of the memoization process, but it does seem a bit expensive and could potentially cause performance problems by doing so. I have to assume there's some way of getting the new node's parent before it actually 'exists' on the page, but Slate being what it is I have no idea how to even approach it. 

I'll flip this back to WIP until I have a solution.

RachelDau (dismissed):
Looks good to me and working in all browsers.

PR #2072, opened 2023-03-20
WIP Dev/31 - Taconite

### LINKED ISSUE #1181: Need an excerpt / aside / box chunk
Two examples seen of authors attempting to visually distinguish content:

Screen Shot 2020-03-12 at 8 43 04 AM
(Done with the HTML chunk)

Another example, using a Code chunk:

image

I had started working on an excerpt type of chunk but didn't finish. But I think these examples show the need for it.

I can see different styles being available. I also see the need to use this not only in the above examples but for things like quoting a portion of a book or poem (for example). So a citation attribute is needed.

Here's a quick draft of what this might look like:

image

### LINKED ISSUE #1575: suggestion: user search for sharing, omit users already shared
This is just polish, but it'd be nice if the user search omitted users that the module has already been shared with.

I can think of a couple approaches
* pass the list of users to omit to the api call
* filter out shared users on the client after search results return (maybe in the reducer)
* add a property to the users passed to the search results that indicate they are already shared with

### LINKED ISSUE #2004: Make community modules previewable.
Currently it's not possible for users to see what a community module looks like without making a copy of it first. Allow logged in users to preview (or even just standard student view) community modules.

### LINKED ISSUE #2021: Quality of life improvements for collections.
List may grow, but immediate concerns:

- When adding multiple modules to a collection, display a 'everything worked' dialog after closing the collection management window. There's nothing that makes it apparent that the modules have actually been added to the collection.
- When creating a new collection, scroll the page so that the new collection is visible.

### LINKED ISSUE #2030: Numeric answer validation doesn't detect shorter non-numeric strings
When entering a non-numeric answer for a numeric question, a form validity element should show saying 'Not a valid numeric value' when trying to leave focus. This seems to work, but only for strings with at least 7 characters. 

This also still allows saving which results in a blank screen error when opening the viewer, but this part should be fixed with #2027
numeric

### LINKED ISSUE #2065: Make question banks collapsible.
Question banks can get very vertically large very easily.

Add some means of collapsing/hiding all of the questions/nested banks in a question bank.

### LINKED ISSUE #2066: Prevent certain node types from being nested within certain other node types.

### LINKED ISSUE #2084: Remove delay between dialog displays when bulk adding to collection.
See #2021.

There's a brief flash of the screen between adding multiple modules to a collection and displaying the success message. See below:

Ideally there should be no delay between the dialogs, hopefully the fix to this should be a relatively quick and easy change to a reducer function somewhere.

### LINKED ISSUE #2085: Add a page to the visual editor tutorial explaining the new Excerpt node type.

### LINKED ISSUE #2088: Pressing the 'up' key in a text node moves the cursor to the beginning of the previous node, not up by one line.
See below:

This isn't exactly show-stopping but it is extremely annoying. It seems to affect any text-based node. Text, List, etc.

### LINKED ISSUE #2095: Excerpt bugs
Problems identified so far:
- If an excerpt contains any text node with text in it and backspace is pressed at the beginning of a citation, the citation's text is moved to the end of that text node. The excerpt is otherwise fine.
  - This might be okay. The behavior is a little weird, but I could see it making sense.
- If an excerpt only contains an empty text node and backspace is pressed at the beginning of a citation, the empty text node is deleted and the citation's text is dumped into the excerpt. The excerpt is not recoverable.
- If an excerpt's citation is empty and backspace is pressed at the beginning of a heading/title node below the excerpt, the editor crashes.
- If an excerpt's citation is not empty and backspace is pressed at the beginning of any text node below the excerpt, the text is moved into the citation. If the enter/return key is pressed anywhere in the citation, the text to the right of the cursor disappears. The excerpt is otherwise fine.
  - I'm not sure what exactly is happening here, but I suspect a new 'line' is being created which is incompatible with the way citations are expected to work. Maybe prevent the enter/return key from doing anything in the citation? Ideally this would take any text in the citation past the cursor and create a new text node underneath the excerpt to insert it into.
- If an excerpt's citation is empty and delete is pressed in the citation, the node below the excerpt is deleted. If this node is a text node, everything seems to keep working. If this node is not a text node (i.e. a figure, iframe, etc.) the editor crashes.

FrenjaminBanklin:
Believe it or not, [https://en.wikipedia.org/wiki/Taconite](https://en.wikipedia.org/wiki/Taconite).

Closes #1181.
Closes #1575.
Closes #2004.
Closes #2021.
Closes #2030.
Closes #2065.
Closes #2066.
Closes #2084.
Closes #2085.
Closes #2088.
Closes #2095.

This release:
 - Removes code meant to solve a keyboard navigation error in Firefox but ended up propagating that error to other browsers.
 - Automatically scrolls to make newly created collections in the dashboard visible, and shows a success message when adding multiple modules to a collection in a bulk operation.
 - Makes Question and QuestionBank nodes collapsible.
 - Adds the Excerpt node type.
 - Enables any node type to define other node types that can not be children of that node type. This is essentially to prevent content authors from nesting nodes infinitely.
   - Question and Excerpt nodes can not contain Question, QuestionBank or Excerpt nodes.
 - Omits users from the 'Find Users to Share With' dialog who already have some level of access to a module.
 - Allows any logged-in user to preview any module reachable from the Module Library.
 - Adds a tooltip to the editor for questions with a numeric/free-input answer type explaining what constitutes valid answers.
 - Bumps `webpack` version to 5.76.0.
 - Bumps `xml2js` version to 0.5.0.


Viggie1:
All the issues closed in this dev branch seem to be working on QA, LGTM

FrenjaminBanklin:
This has been running in our QA environment for about a week and we haven't found any issues. I'm going to go ahead and start finalizing it for release.

PR #2073, opened 2023-03-23
Adds a split-run (A/B test) option to the module embedder.

### LINKED ISSUE #1453: A / B Testing / Conditional LTI Module Assignment
Feature request to split students into groups when starting an Obo LTI launch. So for example, one LTI assignment link could conditionally send students to one of two modules.

Either students are randomly selected to a module, optionally that decision is stored, or hardcoded lists are provided to determine which module to send students to.

This feature may not be implemented natively in Obojobo, but still marking an issue here for when our system supports this.

FrenjaminBanklin:
Closes #1453.

Allows instructors to embed either a single module (current behavior) or a split-run, which will send students to one of the two selected modules based on the timestamp of their user record creation in Obojobo.


RachelDau (commented):
It looks good and is working when I tested it out for both students and instructors. 

My one question is since the module selection is based on when the user was created, should a particular user always go to the same module in a split. In my test, the same user did not always go to the same module when launching from the lti page. I am not sure if this is an issue. Here is the code I was looking at:
`const chooseSplitRunDraft = (user, draftA, draftB) => (user.createdAt % 2 === 0 ? draftA : draftB)`

can draftA and draftB switch on different runs? Will this impact students who I assume should always be taken to the same module once they begin from their webcourses.

FrenjaminBanklin:
Since the module presented in a split-run embed depends on the user's creation timestamp in the database, they should be getting the same one every time. I'm fairly certain that code is working properly - the thing that's changing sporadically appears to be `user.createdAt`. I'm going to investigate and see if I can figure out why.

FrenjaminBanklin:
It looks like there were competing expectations for when the individual validators would be running. As part of the LTI launch/user login process `req.currentUser` was being set with the current timestamp as the `createdAt` value, which of course is not the same as the user's `created_at` value. Modified the User model to always update `createdAt` when running `saveOrCreate`, seems to have made results more consistent.

RachelDau (approved):
looks good to me, what a cool feature

PR #2074, opened 2023-04-06
Add optional read-only mode for copied modules and optional update auto-sync for read-only copies.

### LINKED ISSUE #2005: Keep track of when public modules are copied.
~~Currently there's no real way of knowing when somebody makes a copy of a module in the community collection. This may be useful to keep track of, and potentially down the road add a feature to potentially make it possible to overwrite a copy if updates are made to the community collection version of a module.~~

~~Possible extra/alternate ending - make it so that copies of community collection modules are read-only. This would enable instructors to embed a copy of a community module that they haven't changed but will allow them to check assessment stats.~~

It turns out metadata has been generated any time a module is copied that enables tracking down copies of modules, though this data isn't exposed anywhere outside of the database. I don't think it's super important to do so, but the ability to make read-only copies of modules that can be synced with the original at will might still be useful.

FrenjaminBanklin:
Closes #2005.

It turns out tracking copies isn't a problem because metadata has always been generated for that purpose.

Adds support for optionally making copies read-only with the ability to automatically synchronize updates from the original. This will allow users to have copies of modules that they can see assessment statistics for while also allowing them to keep their copy up-to-date with any content changes.


RachelDau (commented):
Everything appears to be working when testing. I looked through the code and it seemed good too, but there is alot of code, and I didn't quite understand everything so another reviewer should also review to be sure.

One thing I noticed is the synchronize button appears on a copied module read-only when you click on it from the dashboard, however if you select `more` and open up the full module options it no longer appears. I do not know if it should appear in module options or how hard it would be to add.

RachelDau (dismissed):
looks good to me

FrenjaminBanklin:
It just occurred to me that there's a 'Make a Copy' option in the document editor. There should probably also be a 'Make a Read-Only Copy' option adjacent to that one.

Setting this back to a draft until I can do that.

Viggie1 (approved):
Everything seems to be working on my end, LGTM.

PR #2078, opened 2023-05-12
Add single-question pacing option to assessments, saves assessment progress to enable resuming.

### LINKED ISSUE #2035: Add a new 'progress bar' rubric option.
Official name could use some work, but the idea is that an additional option would be added to the assessment rubric that would enable a new kind of assessment grading scheme.

In this new grading scheme some percentage 'progress' is awarded per question answered correctly, and upon reaching 100% the student would receive full credit for the assessment and it would conclude immediately. Answering a question incorrectly would remove some percentage of 'progress' from the current total. Answering all questions before achieving a 100% score would submit the final percentage as the score for the assessment, or at least use that percentage in all further rubric calculations.

The percentage awarded/lost per correct/incorrect answer should be individually customizable. Ideally using this option would also add a progress bar to the viewer side so students would be aware of their progress. It may make more sense if this option is in the assessment three-dot menu rather than the rubric, since the final score could still be subject to rubric modifications.

FrenjaminBanklin:
Closes #2035.

Branch name is misleading - 'progress bar' assessments are (probably) currently not possible.

This PR adds a pacing option for assessments, defaulting to 'all questions visible simultaneously' but optionally allowing for authors to present assessments one question at a time.

Assessment states are now saved, pre-filling selected answers if a student resumes an assessment attempt. In single-question pacing, the assessment will also resume at the last question rather than at the beginning.


RachelDau:
Picking up where you left off works well and saves the answers selected. I am noticing some things up with the one at a time question feature. It will be working fine and then I will get stuck on a page because the next or submit button just stops working when clicked. I included the error screenshot from pressing the next button.

Also it is a very abrupt transition between questions although this is a style preference and not critical to function. 
Screen Shot 2023-05-16 at 1 49 19 PM

FrenjaminBanklin:
Setting this back to WIP while I debug the discovered issues with single-question pacing.

FrenjaminBanklin:
@RachelDau and @Viggie1 - anything else you may have noticed prior to experiencing the errors would go a long way to helping me figure out how to reproduce them.

FrenjaminBanklin:
It looks like this can be reliably triggered by completing an assessment and then re-attempting that assessment immediately, then trying to answer a couple of questions. There's probably a counter somewhere not being reset properly. It looks like the assessment is also in a weird quasi-review mode, which displays the selected answer and indicates if it's correct or not. Depending on the assessment settings this may also show what the correct answer is regardless of how many attempts are left, but I need to experiment with that.

I'm also noticing that the screen briefly flashes a different dialog state between questions in single-question pacing mode, which is far from desirable. I'll have to try figuring that out, as well.

FrenjaminBanklin:
Closing this in favor of #2106 and some future second PR to address one-question-at-a-time pacing.

PR #2080, opened 2023-05-15
Issue/2021 quality of life improvements collections

Viggie1:
Implemented success dialog when adding multiple modules to a collection, scrolling to newest created collection, and fixed '# Modules Selected' not zeroing out correctly.


packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -199,8 +209,18 @@ const renderModuleManageCollectionsDialog = (props, extension) => (
...
+		title=""
+		collections={props.myCollections}
+		selectedModules={props.selectedModules}
+		bulkAddModulesToCollection={props.bulkAddModulesToCollection}
+		onClose={extension.onClose}
+	/>
+)
+
+const renderBulkSuccessDialog = props => (
+	<BulkSuccessDialog
```
FrenjaminBanklin, line 223: The `BulkSuccessDialog` component doesn't appear to be using `collections`, `selectedModules` or `bulkAddModulesToCollection`. If they're not necessary to that component, it's not necessary (and cleaner) to not provide them to that component as props. I'd remove them here.

packages/app/obojobo-repository/shared/components/dashboard.jsx
```diff
@@ -647,7 +674,10 @@ function Dashboard(props) {
 				<div className="repository--item-list--collection">
 					<div className="repository--item-list--collection--item-wrapper">
 						<div className="repository--item-list--row">
-							<div className="repository--item-list--collection--item--multi-wrapper">
+							<div
+								className="repository--item-list--collection--item--multi-wrapper"
+								ref={newCollection}
```
FrenjaminBanklin, line 679: Setting the collections container as the ref instead of the new collection itself seems like it shouldn't work, but I can see that somehow it does. I'm willing to let it ride, but we may have to go back and adjust this to add the ref to a specific collection component as it's being generated inside `renderCollections`.

packages/app/obojobo-repository/shared/components/dashboard.test.js
```diff
@@ -1059,6 +1070,75 @@ describe('Dashboard', () => {
 		component.unmount()
 	})
 
+	test('updates selected module count properly when all modules are deselected', () => {
```
FrenjaminBanklin, line 1073: The indices chosen here (`standardModules[0]`, `moduleComponents[4]`, etc.) are a bit magical - may be prudent to add a comment or two near those explaining why we're using those numbers in particular for the benefit of future devs.

yarn.lock
```diff
@@ -4550,9 +4550,9 @@ camelize@^1.0.0:
   integrity sha1-FkpUg+Yw+kMh5a8HAg5TGDGyYJs=
 
 caniuse-lite@^1.0.30000844, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001181:
-  version "1.0.30001204"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001204.tgz#256c85709a348ec4d175e847a3b515c66e79f2aa"
-  integrity sha512-JUdjWpcxfJ9IPamy2f5JaRDCaqJOxDzOSKtbdx4rH9VivMd1vIzoPumsJa9LoMIi4Fx2BV2KZOxWhNkBjaYivQ==
+  version "1.0.30001469"
```
FrenjaminBanklin, line 4553: Looks like you may have used `npm` to do something instead of `yarn` at some point, though I'm not sure why that would change yarn's lockfile. I would probably undo the changes to this file.

FrenjaminBanklin (changes requested):
This looks pretty good overall. There are a couple of minor nitpicks detailed below, but the main thing I see is that there isn't a test file for the bulk success dialog component. Most likely it's showing as 'covered' because it's being imported and used as-is in another component that's being tested.

If it's a quick change to make the `BulkSuccessDialog` component into a generic `MessageDialog` component, you could do that as part of this PR. If not, spin up a new issue for it once this PR is merged.

packages/app/obojobo-repository/shared/components/message-dialog.jsx
```diff
@@ -0,0 +1,19 @@
+require('./message-dialog.scss')
+
+const React = require('react')
+const Button = require('./button')
+
+const MessageDialog = props => {
+	return (
+		<div className="bulk-success-dialog">
```
FrenjaminBanklin, line 8: Nitpicking - change this class name from 'bulk-success-dialog' to 'message-dialog' to reflect the change to the component's name. Same change in the adjacent .scss file.

FrenjaminBanklin (changes requested):
One super minor change,  otherwise this looks like it's good to go.

FrenjaminBanklin (approved):
There's a momentary delay between the bulk add window and the success message where the dashboard is suddenly visible for a second. Ideally the transition between states would be more fluid, but I don't think it's enough of an issue to delay merging this in.

Perhaps we can spin off another issue to look into making the transition seamless. Calling this good to go for now.

PR #2082, opened 2023-05-24
Reset element IDs in copied modules.

### LINKED ISSUE #2081: When copying a module, re-ID all nodes.
Currently when a module is copied, the draft ID (and draft content ID) are changed, but the document (aside from the title) otherwise remains the same.

Find a way to crawl the document and give each node a new ID when copying a module.

FrenjaminBanklin:
Closes #2081.

When making a copy of a module, all element IDs following the UUID format will be replaced. References to the old IDs in button actions etc. will also be updated with the corresponding new ID.


packages/app/obojobo-repository/server/routes/api.js
```diff
@@ -186,10 +188,39 @@ router
...
 			draftObject.content.title = newTitle
-			const newDraft = await Draft.createWithContent(userId, draftObject)
+
+			// convert the object to a JSON string so we can swap out all the old IDs with the new ones
+			let draftString = JSON.stringify(draftObject)
+
+			// globally replace each old ID with the equivalent new ID
+			// this should replace node IDs as well as action trigger references to those node IDs
+			for (const [oldId, newId] of Object.entries(idsForChange)) {
+				// this works, but there may be a more efficient way of doing it
```
mbusch3, line 216: This is a very elegant solution... Changing keys while retaining values is a pain, while string replacing is simple and relatively fast. The only real concern is making sure you're replacing globally instead of the first instance only, which you've done below AND made sure was included in your test case.

mbusch3 (dismissed):
I don't know enough about the existing code to know if this is a problem or not. But in addition to ids that were changed, in the parent object, this also changes the 'rev', 'draftId', and 'contentId'. Do these reference other objects that need to be saved to the database, or does Draft.createWithContent cover that? The front end seemed to work fine when loading and editing both the original and the copy.

FrenjaminBanklin:
A copied module is a new module, so `draftId` and `contentId` being new is expected. I'm not exactly sure where `rev` is set, but my guess would be that it's also specific to an individual draft or revision so it's probably fine that it also changes.

FrenjaminBanklin:
This was previously reviewed and everything still appears to be working correctly after modernizing from the current dev branch. I'm going to go ahead and call it good.

PR #2086, opened 2023-06-08
Adds onClear function for the bulk add modules dialog to prevent flicker

### LINKED ISSUE #2084: Remove delay between dialog displays when bulk adding to collection.

mbusch3:
Resolves #2084


FrenjaminBanklin (approved):
Does the job and looks good to me.

PR #2077, opened 2023-05-11
Make question banks collapsible.

FrenjaminBanklin:
Closes #2065.

Adds collapse/expand buttons to Question and QuestionBank nodes.

QuestionBank nodes also now have buttons to quickly expand/collapse all child Question or QuestionBank nodes.

PR #2087, opened 2023-06-09
Collapsible code review

mbusch3:
Code review of #2077 along with unnecessary styling changes and updated unit tests to match.


FrenjaminBanklin (changes requested):
In general I like the idea of making the labels themselves clickable to expand a collapsed question/question bank, but it looks like you may need to play with the specificity of the cursor change a bit more.

See the attached gif below. The cursor does not change to a pointer when hovering over a collapsed question that is not inside a question bank, and the cursor is a pointer everywhere when hovering over a question inside a question bank whether the question is collapsed or not.

FrenjaminBanklin (approved):
This looks like it's working well.

PR #2089, opened 2023-06-12
Removes code that overrides up arrow functionality

mbusch3:
Addresses Issue #2088 by removing offending code and test cases.


FrenjaminBanklin (approved):
This solves the problem in Chrome and Safari, at least. I'll call it good enough to merge but leave it open for the first half of tomorrow in case you manage to find a long-term solution that works for all of the browsers. Otherwise, this is good enough for the next release.

FrenjaminBanklin:
It looks like the specific problem in Firefox is that when the cursor is at the top of a text node, pressing the 'up' key sends the cursor to the beginning of the entire page rather than to the beginning of the next node in that direction. Moving around within a text block seems to be okay.

PR #2096, opened 2023-06-23
Resolves crashes when moving text upwards in a document into an excerpt citation.

### LINKED ISSUE #2095: Excerpt bugs

FrenjaminBanklin:
Closes #2095.

Simplifies excerpt citations to allow text to be moved into them correctly, resolving what was previously a total meltdown in the visual editor.

Also snuck in a minor change to the name of the component in insert lists, and a corresponding update to the relevant page in the visual editor tutorial module template.


mbusch3:
I have been unable to produce the complete system crash that was the primary issue here. Almost all of the behavior is consistent and expected. The issues below could probably be broken off into their own issues and resolved later, since they aren't of major concern.

Note: I wasn't able to test the delete key because it's not on my keyboard. I tried using ^D as it's the Mac shortcut for delete, but that was being used to toggle on/off strikethrough, and I wasn't able to remap the 'delete' shortcut on my machine.

Good behavior:
- When there is text in the citation, selection and hitting backspace inside the citation work as expected
- When there is text in the citation, hitting backspace from the element below appends its value to the end of the citation as expected

Unexpected behavior:
- When there is no text in the citation, hitting backspace inside the citation does nothing. This should navigate up into the bottom element of the excerpt content without deleting the citation to be consistent with other keyboard navigation behavior.
- When there is no text in the citation, hitting backspace from the element below skips over the citation and focuses the cursor in the excerpt content. If there was text in the element below, it does not append to the excerpt content. I'm not sure whether we would prefer the text be appended to the citation or to the excerpt content, but when there is no text, skipping over the citation feels inconsistent.
- When the excerpt is the first element on screen, hitting backspace from the top element of the excerpt content moves the cursor down into the citation. It should probably stay in the same place.
- Every once in a while, when trying to use backspace to append text content into the citation, the entire element is inserted into the citation instead of just the text (see image). This seems to happen (a) with header elements, and (b) after using the 'Undo' feature. This can always be undone and does not cause a system crash.
Screen Shot 2023-06-26 at 9 16 48 AM

mbusch3 (approved):

PR #2098, opened 2023-06-27
Updates Node to version 18.

### LINKED ISSUE #2013: Upgrade Node to version 18.

FrenjaminBanklin:
Resolves #2013.

Made minimal changes to locked package versions to support Node 18. Updated Node version requirements in relevant package.json files, fixed new undefined error message in tests.

More specifically, adjusted the Node engine version requirements in all `package.json` files and removed the following dependencies from `yarn.lock` so they would be re-locked at more recent/compatible versions:

- canvas@^2.6.1
- webpack@5.30.0
- postcss@


Viggie1 (approved):
I have loaded this branch onto my machine and tested it locally. Everything worked, LGTM.

PR #2102, opened 2023-07-18
Dev/32 - Pigeonite

### LINKED ISSUE #1869: Assessment Stats Dialog Phase 2
This is a follow up to work done on #1864 . There are some refactors and improvements still to do to the assessment stats feature:

* Merge AssessmentStatsSearchControls and AssessmentStatsFilterControls into one component, and have all filtering happen there
* Change search by text input to a hybrid text-input/select, which shows all of the possible values
* Change First / Last name to a single field, which searches both fields
* Set the min date of the date range inputs to the first day of the earliest record

### LINKED ISSUE #2005: Keep track of when public modules are copied.

### LINKED ISSUE #2013: Upgrade Node to version 18.

### LINKED ISSUE #2081: When copying a module, re-ID all nodes.

### LINKED ISSUE #2104: Update `stylelint` dependency from version 13 to version 15.
See the `dependabot` PR, #2099.

There are many other necessary changes beyond just updating the version number and any sub-dependencies.

There will have to be a whole landslide of CSS changes at minimum but I also suspect whole dependencies will have to be abandoned in favor of more modern equivalents.

FrenjaminBanklin:
https://en.wikipedia.org/wiki/Pigeonite

Closes #1869.
Closes #2005.
Closes #2013.
Closes #2081.
Closes #2104.

This release:
 - Adds read-only copies, which can not be modified but allow owners to synchronize changes from the original. The existing `drafts_metadata` table is used to track read-only status.
 - Changes the ID of every content node in a document that is created as a copy of another document, including references to those IDs in actions.
 - Improves the Assessment Stats dialog in the module dashboard.
   - Filtering and searching is now possible with a single control, which also functions as both a text input and a dropdown selector for easier searching. First and last name are combined for searching purposes, and date ranges are constrained to available results for convenience.
 - Bumps the version of the Node engine from 14 to 18.
   - This required re-locking the `canvas`, `webpack` and `postcss` dependencies (and their subdependencies) at higher versions for compatibility.
 - Bumps the `word-wrap` dependency version from `1.2.3` to `1.2.4`.
 - Bumps the `semver` dependency version from `5.7.1` to `5.7.2`.
 - Bumps the `stylelint` dependency version from `13.12.0` to `15.10.0` and adjusted associated dependencies and Stylelint rule configurations.


mbusch3 (commented):
Reviewed each of the new and changed files. Nothing jumped out at me. Still need to run tests on all of the pigeonite functionality. I've made new modules with read-only copies, but haven't tested the new assessment data searching and such.

PR #2106, opened 2023-07-26
Saves assessment attempt responses, recalls them when resuming an unfinished attempt.

### LINKED ISSUE #2105: Make resuming an assessment attempt actually resume the attempt instead of just starting it over.
Currently none of the progress made in an assessment attempt is actually saved - any answers made to questions will evaporate if the attempt isn't finished in a single session.

FrenjaminBanklin:
Closes #2105.

Adjusted assessment store/state management to support saving assessment progress when providing a response to a question.

Assessment response states are now saved and recalled properly when resuming an unfinished assessment attempt.


RachelDau (dismissed):
Testing this out and it is working great. I tried in on multiple browsers and with all types of questions. This looks good to me :)

FrenjaminBanklin:
Technically this should be approved again, but I've tested it after merging dev/33 and nothing has changed so I'll just honor the old reviews and call it good.

PR #2110, opened 2023-09-18
Added bookmark to edge tests

Viggie1:
Added bookmark/dogear excerpt option to the following excerpt styles:
-filled box
-bordered box
-card 
-white paper
-modern paper
-light yellow paper
-dark yellow paper
-aged paper


FrenjaminBanklin (approved):
We discovered some things in the pursuit of this that I think are worth addressing in another issue, but this looks good considering the single mockup we had to go by.

I did technically contribute to this myself however, so another pair of eyes would be nice.

RachelDau (approved):
It looks good and works for me on different browsers. I see this doesn't line up exactly cosmetically but its not really noticeable and I think it was difficult to do so I still approve this PR
image

PR #2116, opened 2023-10-10
WIP: Variables and experimental features.

FrenjaminBanklin:
**THIS IS A WORK IN PROGRESS. THIS PULL REQUEST WAS CREATED TO MORE EASILY COMPARE HISTORICAL CHANGES AGAINST CURRENT CODE.**

Adds variable creation and substitution and experimental feature flags.


FrenjaminBanklin:
This is just about done and working, but there are a couple of potentially serious problems that we probably need to address before we can ship this.

These problems only really surface when using the `{{$varname}}` shorthand. Substitution appears to be working properly in all cases when using the `{{$ownerid:varname}}` format, but I don't think we can reliably expect content authors to be mindful of this and it kind of goes against the whole purpose of being able to define variables in multiple places using the same name.

Paging @zachberry for any ideas for approaching the things listed in 'Problems' in the PR description above. Currently it doesn't look like there is any way of circumventing the way Obojobo handles questions in assessment attempts and during assessment review in order to make the questions' full node ancestor hierarchies available to the variable util. I'm considering trying to find some way of attaching each question's full ancestor tree on the server side and then somehow using that information on the viewer side, but even in the best case scenario the solution is going to be extremely hacky.

zachberry:
I probably need to re-familiarize myself a bit more, but some initial top-of-my-head thoughts:

* Could variable substitution happen server-side for assessments? When the server is choosing which questions to give to the client I think you'd have the whole tree in memory at that point. The server tree objects are a bit different than OboModel but maybe they could still be used. Then when the client gets back the question it doesn't get back "What is {{a}}?", it gets back "What is 0.123?".
* Or, determine the value of all variables on the initial viewer page load and return them along with the viewer JSON. They could be connected to the ID of the node they're used in. So for example, if you have a document like this (this example deliberately uses both shorthand and longhand):

```
...
   Page (id=p1, variables: studentName=chooseOne('Dick', 'Jane'))
      TextChunk (id=tc1)
         Text (id=t1)
            textGroup: "Imagine a student named {{ studentName }}. When {{ p1:studentName }} drives 10 miles..."
```

The server could return the following when requesting a document (I'm not tied to this exact structure, just trying to convey the idea):

```json
{
  root: { //...the whole obo document...// },
  variables: [
    {
      name: "studentName",
      value: "Jane",
      owner: "p1",
      refs: ["t1"]
    }
  ]
}
```

So the server comes back with a variable value it chose (Jane), where the variable is defined (the "owner", which is "p1", the Page) and finally anywhere the variable is used (the "refs", which is "t1", the TextChunk that uses it).

(You would ideally want to further process the `variables` data client-side to make it more useful to quickly look up - maybe make a data structure like this:)

```js
const variableValuesByOboNodeId = {
  p1: {
    studentName: "Jane"
  },
  t1: {
    studentName: "Jane"
  }
}
```

That way when the parser/viewer comes to `{{ studentName }}` or `{{ p1:studentName}}` you would have enough information to find what its value should be.

If this was done for everything in the document then you would be fine if you're missing part of the node structure, such as in Assessment, because that `variables` response gives you everything you need to connect the dots.

Thoughts? I might be missing something obvious since it's been a minute since I looked at this stuff.

FrenjaminBanklin:
My knee-jerk reaction to performing variable substitutions server-side but just for assessments is that we may as well just perform substitutions server-side for the whole document. There was a lot of code that seemed like it wasn't being used that I just got rid of, I'm not sure if any of it would have been supportive of this approach or if it was just left over from some unrelated experimentation. But then that seems like it would require some pretty significant rearchitecting of not just the way variables work but the way the whole viewer backend works. I'm also not sure if moving substitutions out of the client side completely would make it harder to use variables in nav links - or since the substitution would be done already if the nav links would just have the correct text to begin with.

I can look into limiting the server-side pre-baked substitutions to just everything downstream of the Assessment. I'd taken a really tentative step in this direction before but aborted because I couldn't see any obvious convenience tools for traversing a draft tree in reverse (there's a `getChildNodeById` method, but not a `getParentNode` method, for example - so that's something we'll probably want to add to make our lives easier) or for easily checking a node's text for the variable string once it's objectified (though admittedly I didn't experiment much with this). It'll be a bit of a hack to have substitution happening two different ways in two different places, but it could just be a 'this works for now, standardize on one later' kind of deal.

FrenjaminBanklin:
The solution I ended up coming up with was a bit of a hybrid. When an attempt is started, each chosen question will have all of its descendent nodes checked for variables. After all of the variables used anywhere in the question have been aggregated, for each variable all of the question's ancestor nodes will be checked until the closest one that owns a variable of the same name. Once a match is found, a reference is stored as part of the `attempt.state.chosen` payload in a new `varRef` key so that it can be referenced when resuming or reviewing an attempt.

Luckily this only has to happen the one time, and then whenever `attempt.state.chosen` is used to pull all of the questions out of a draft, a separate step can recurse through all of the child nodes for each chosen question if a `varRef` exists for that question, and whenever a string matching the given variable is found it will just be replaced with the long form syntax.

So if a question contains any text nodes with a `{{$qb}}`, and that question's parent QuestionBank with an ID of 'qb-id' is the nearest node with a 'qb' variable defined, then before that question reaches the client for rendering all occurrences of `{{$qb}}` within the question will be changed to `{{$qb-id:qb}}`. At that point the existing tools on the client side should handle it.

PR #2126, opened 2024-03-08
Updates LTI config URLs to use https address instead of http address.

### LINKED ISSUE #2125: Update all `http` URLs to the IMS Global LTI rules to `https`.
For example, all instances of `http://www.imsglobal.org` should become `https://www.imsglobal.org`.

FrenjaminBanklin:
Closes #2125.


FrenjaminBanklin:
Tests etc. pass, skipping review for this one to fast-track an alpha tag for testing in a QA environment.

PR #2139, opened 2025-12-17
Modifies Materia OboNode server-side code to work with the upcoming Materia release.

FrenjaminBanklin:
Materia's upcoming release removes support for LTI 1.1 in favor of LTI 1.3, requiring changes on the Obojobo end to have it act as a basic LTI 1.3 platform.

This update also pins a more recent commit of the upstream `ims-lti` repository to add newly required headers to outcomes requests sent to an LMS upon completing a module assessment.

This requires a few additional environment variables:
- `MATERIA_LTI_VERSION`: Either `1.1` or `1.3`.
- `OBO_LTI_CLIENTID`: This should be an alphanumeric string value determined by the platform (Obojobo) that needs to be matched in a corresponding LTI registration in any relevant tools. It can be an arbitrary alphanumeric string.
- `OBO_LTI_UUID`: This is a UUID value that is generated by the Materia LTI registration.
- `OBO_JWT_KEY`: This is an alphanumeric string value.
- `OBO_PRIVATE_RSA_KEY`: This is an RSA key generated by a command like `openssl genrsa -out private.pem 2048`. Ideally condensed to a single string with newline characters and bookends preserved.
- `OBO_LTI_DEPLOYMENTID`: This is a UUID value that is generated by the Materia LTI registration.


FrenjaminBanklin:
This has been running in production for us since the new year. Not really enough to merit a full release.

